PR #1, opened 2019-08-20
Make copy html assets when build time

mog422:


pypy-vrc:
It is intended. sorry.

PR #2, opened 2019-08-20
Show name color by trust level when display name

mog422:
image


pypy-vrc:
I didn't color the nickname by rating because I thought it was no longer meaningful. But you made a request, I'll add that as an option. Thank you.

PR #5, opened 2019-09-24
replaced vrchat icon with vrcx icon

Fusl:
because sometimes it's very confusing when two applications (vrchat and vrcx) have the same icon in the taskbar and it takes you two clicks to get the right one


pypy-vrc:
thank you :)

PR #12, opened 2019-10-13
Use invites without worldId for sending messages

GotoFinal:
VRChat probably disabled message type, so it needs to be done using invites instead.

Also it would be nice to think about splitting this code base, its pretty hard to work on single file like that

### DIFF
--- a/html/app.js
+++ b/html/app.js
@@ -6159,10 +6159,10 @@ if (window.CefSharp) {
 								instance.inputValue) {
 								API.sendNotification({
 									receiverUserId: D.id,
-									type: 'message',
+									type: 'invite',
 									message: instance.inputValue,
 									seen: false,
-									details: '{}'
+									details: {worldId: '', worldName: instance.inputValue}
 								}).then((args) => {
 									this.$message('Message sent');
 									return args;


Fusl:
I'm personally against merging this as messages sent via instance invites is one way to easily tell apart a modded client from a standard client on the server side and people using VRCX may be mistaken for using a modded client.

GotoFinal:
How is that different from sending request to inactive endpoint that was working before and was never available in any feature in game? Clients were using old message request too, but they closed it as it was unused, so people moved to another way of sending messages.

Fusl:
You're right, it's no different from that. What I should have said is that I'm against merging this without adding an explicit alert box that sending a message via VRChat may increase your chances of getting banned from the platform.

Murkur:
ive seen clients use this method instead saying
Join me in (CUSTOM MESSAGE)
its pretty cool would actually like to have VRCX say it aswell
since its kinda weird when it says (Join me in HIYA)

GotoFinal:
@Murkur This method dos include "Join me in..." part, when did you see clients that can send such messages to normal users? (as some clients might include custom communication system between users of same client) As if there is some other method I could try to use it, but I just used the simplest working one I found.

 @Fusl I don't really think this is in any way bad. we are not doing anything miscellaneous and we are not touching the game so it does not break the rules.
I will just leave that PR for now, and someone can always adjust this for own needs

Murkur:
> @Murkur This method dos include "Join me in..." part, when did you see clients that can send such messages to normal users? (as some clients might include custom communication system between users of same client) As if there is some other method I could try to use it, but I just used the simplest working one I found.
> 
> @Fusl I don't really think this is in any way bad. we are not doing anything miscellaneous and we are not touching the game so it does not break the rules.
> I will just leave that PR for now, and someone can always adjust this for own needs

The clients I know off still say (Join me in) but instead they tack on (Custom message) ahead of user input to make it easier to understand that its not an invite
so example being
Join me in (CUSTOM MESSAGE) Hiyaaa

Fusl:
@GotoFinal Can you replace that part
```
message: instance.inputValue,
```
with
```
message: '',
```
and add a warning message for users that this feature uses the invite system in a way that it was not designed for and may get them banned so using it at their own risk?

Murkur:
> message: '',

the changes of getting banned for this is very slim but still possible as vrchat doesnt log invites but they do log information about other endpoints like a friendrequest as adding people boosts your rank within vrchat i got a test account banned for a month by spamming a friendrequest endpoint with vrcx to one of my accounts and the ban reason was "client usage" so yes a warning should be included

Fusl:
@pypy-vrc Can we consider adding a checkbox in the settings/more section of the app to enable dangerous features like this and clearly describe that using such features may, and probably will, result in a ban of the account? This way, we can merge this feature in a way that its only reachable and properly usable once this box has been checked.

Murkur:
I agree with what @Fusl said as chances of getting banned is slim to non but should still contain the warning for people that dont know any better or understand how vrchats API works

pypy-vrc:
This is an abuse of the invite message. If someone could spamming to someone with this and it became BAN, the cause is VRCX, and I will be sad. 😥

Arion-Kun:
Hey, could someone look at the code to see how the update changed some things. 

Iv tried this code only to have a white screen while loading it up. It's the old way to make it function. I suspect its due to the "+instance.inputValue" might have been changed but I don't know any .js I just learn some when I need to make it function.

This is the current code iv tried and makes the app not work:
`{receiverUserId:t.id,type:"invite",message:r.inputValue,seen:!1,details:{worldId: '', worldName: '-| ' + instance.inputValue}`

Also the app.js has been written in one line making editing and work on it rather difficult.

Murkur:
@Arion-Kun I just reverted it back to the original 
and unminified the JS

```
                callback: (e, r) => {
                    "confirm" === e && r.inputValue && n.sendNotification({
                        receiverUserId: t.id,
                        type: "invite",
                        message: '',
                        seen: false,
                        details: {worldId: '', worldName: '(Custom Message ) ' + r.inputValue}
                    }).then(e => (this.$message("Message sent"), e))
                }
            });
```

EntityReborn:
I would like to see this done as well.
A warning that a ban might happen would be sufficient.

Murkur:
JS has been modifed a slight bit in the newest re;ease if you want invite messaging still you can still acomplish this by unminifing the JS with a website like [Unminify](https://unminify.com/) and then changing line number 43624 to reflect as 


```js
                                    c.sendNotification({ receiverUserId: n.id, type: "invite", message: i.inputValue, seen: !1, details: {worldId: '', worldName: '(Custom Message) ' + i.inputValue} }).then(function (e) {

```
image


screenshot just to see

PR #43, opened 2020-03-27
make vr overlay dark themed

Fusl:
image
image


Murkur:
yeah this would be a great addition ive just been manually adding it myself after each update but finally integrating it would be good

Natsumi-sama:
Added with: https://github.com/pypy-vrc/VRCX/commit/1953098790593df30dfa59f88bda725a97de0d0e

PR #77, opened 2020-09-15
VR SCSS rework and themes

N3X15:
**I haven't tested this in VR yet**, beyond making sure it doesn't crash. Will do that this afternoon after the thunderstorms die off.

This PR does a complete overhaul of the VR notification SCSS, making it use more features of SCSS and using them to make two themes: light and dark.  I've also added some JS to (hopefully) permit switching between the two after reading the user's app settings.

[rant]<br />
I'm going to be honest here, and please bear in mind that I'm not trying to disparage the developers:  The UI for this application is great, but the way it's put together is highly inefficient and redundant.  I have half a mind to do a more thorough overhaul to clean things up and get rid of a lot of dead code and redundancies.  One particular example is `uuidv4()`, which is actually available in the `uuid` package pulled in as a dependency (`node_modules/uuid/v4.js`).  The way `uuidv4()` is implemented right now is incorrect and insecure.  On top of this, there's a lot of polyfills that we don't actually need for Chromium-based applications.
<br />[/rant]

### DIFF
--- a/html/src/index.pug
+++ b/html/src/index.pug
@@ -1058,19 +1058,19 @@ html
                         p(style="font-weight:bold") animate.css
                         pre(style="font-size:12px;white-space:pre-line").
                             The MIT License (MIT)
-                            
+
                             Copyright (c) 2019 Daniel Eden
-                            
+
                             Permission is hereby granted, free of charge, to any person obtaining a copy
                             of this software and associated documentation files (the "Software"), to deal
                             in the Software without restriction, including without limitation the rights
                             to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
                             copies of the Software, and to permit persons to whom the Software is
                             furnished to do so, subject to the following conditions:
-                            
+
                             The above copyright notice and this permission notice shall be included in all
                             copies or substantial portions of the Software.
-                            
+
                             THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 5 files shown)


pypy-vrc:
Thank you for contribution.

I didn't use webpack or package.json to implement VRCX in the first time, so all the codes in one place. It should be clean up. I fully agree with you. but, the uuidv4 implementation in VRCX does not need to be strict. It satisfies the version bit in the uuid specification, and fills the rest randomly, but it is enough to use for VRChat. It simply needs a random string.

You changed the dark mode implementation to a descendant selector rather than the old one, is there a reason you did this? I think this will cause performance problems. that's why I put CSS as a separate file and controlled with the link tag.

Fusl:
> One particular example is `uuidv4()`, which is actually available in the `uuid` package pulled in as a dependency (`node_modules/uuid/v4.js`). The way `uuidv4()` is implemented right now is incorrect and insecure. On top of this, there's a lot of polyfills that we don't actually need for Chromium-based applications.

The way uuidv4 is implemented isn't incorrect but it is indeed cyptographically insecure, it was more or less a copy paste from a random StackOverflow and [my first contributions](https://github.com/pypy-vrc/VRCX/issues/4) to this repository. I avoided the use of Node.js functions and npm package libraries as I didn't know whether CefSharp would support such implementations and VRChat itself doesn't use a cryptographically secure function to generate those either so the time investment in figuring that out was a little over my head at that time.

That said, I agree that lots of the JS/ES, CSS and HTML, generally everything of the internal browser of VRCX should be given a full rewrite with a focus on security, stability and performance.

Natsumi-sama:
Closing for inactivity.

PR #95, opened 2020-10-27
VR overlay notifications dark mode toggle

Natsumi-sama:
Sorry @Fusl for not keeping your original commit, fork too old.

### DIFF
--- a/html/src/vr.js
+++ b/html/src/vr.js
@@ -79,7 +79,7 @@ CefSharp.BindObjectAsync(
         },
         layout: 'topCenter',
         theme: 'relax',
-        timeout: 6000
+        timeout: 3000
     });
 
     Vue.use(ElementUI, {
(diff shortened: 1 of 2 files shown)


html/src/vr.js
```diff
@@ -79,7 +79,7 @@ CefSharp.BindObjectAsync(
         },
         layout: 'topCenter',
         theme: 'relax',
-        timeout: 6000
+        timeout: 3000
```
pypy-vrc, line 82: 👀

Natsumi-sama: lol yea I guess this is a bit of an opinionated change, maybe this should be a setting as well, more settings! 😆

pypy-vrc:
commit merged (8b7f6050e7f0245e37065ad3d8d477fad2350914)

PR #111, opened 2020-12-03
expands avatar favorites from 16 to 25x4

kazu0617:
VRChat Plus announced that amount of avatar favorites expand 25 and 4 categories.
In API, It can be use at now, so change these arguments and expand it.
in this pullrequest doesn't check the owner is plus account or not sry ;)
image
image

### DIFF
--- a/html/src/app.js
+++ b/html/src/app.js
@@ -2602,14 +2602,14 @@ CefSharp.BindObjectAsync(
         }
         // 16 = ['avatars1'] x 16
         this.favoriteAvatarGroups = [];
-        for (var i = 0; i < 1; ++i) {
+        for (var i = 0; i < 4; ++i) {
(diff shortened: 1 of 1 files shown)


kazu0617:
It will be some problems with initial run. check that movement and fix it.
image

kazu0617:
I found how to resolve, but too hard coded... Ready for review.

in api, avatars/favorites must contains tag such as "avatars1" ~ "avatars4". but line 2906@app.js doesn't have any tag, so load only avatars1. if load correctly, must cycle loading.

fwenny:
they unfortunately dont seem to use the tag field.. i tried doing a request with tag = avatars2, but it still only returns avatars2.
there doesnt seem to be an endpoint that supports returning the other avatar groups similar to "avatars/favorites"

PR #218, opened 2021-06-08
update batch files

0xS3GF4VLT:
After pulling and attempting to build the project myself it came to my attention that these batch scripts may be in need of an update. I've compared the build file of the current release to the build I made myself and my conclusion is that these changes are more accurate for the expected input and output.

Also update the 7z batch file to be more inline with the Bandizip batch file. The 7z batch file assumed an exact location for the executable, but the executable should be available as an environmental variable.

Please let me know if you have any issues with these minor adjustments! <3

### DIFF
--- a/make-junction.cmd
+++ b/make-junction.cmd
@@ -1,3 +1,3 @@
-mklink /J "%~dp0\bin\x64\Debug\html" "%~dp0\html\dist"
-mklink /J "%~dp0\bin\x64\Release\html" "%~dp0\html\dist"
+mklink /J "%~dp0\bin\Debug\html" "%~dp0\html\dist"
+mklink /J "%~dp0\bin\Release\html" "%~dp0\html\dist"
 pause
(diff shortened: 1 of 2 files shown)


0xS3GF4VLT:
> All the other build scripts also assume 64-Bit, I don't think that needs to be changed unless I'm missing something. Also 7-Zip isn't in my PATH variable so I assume it wont be for everyone else.

Ah, I hadn't noticed the other build files. Are those even being used?

Also, for me 7z is not installed in the default location so the script failed to execute it without clear messaging.

Natsumi-sama:
Not by me but others may be using them https://github.com/pypy-vrc/VRCX/pull/188

Yea I'm sure there could be a better approach to this.

Natsumi-sama:
pypy removed AnyCPU from the configurations list so now the build should end up in the same location as the scripts. https://github.com/pypy-vrc/VRCX/commit/91a0432cf7fd9c275a0b2977d67bb94257c85d50

Natsumi-sama:
https://github.com/pypy-vrc/VRCX/commit/117fbf84b906fb4a8a5bf2bc4066317588f7649c

PR #222, opened 2021-06-11
More language

GizmoOAO:
Add more language.
But I'm not sure if the sign language needs an icon.


Natsumi-sama:
I think it's better with flag icons because without it the language wont be displayed on the users profile. https://github.com/pypy-vrc/VRCX/commit/248d88506c6f4ed2cd0ca08855ea76001995a545

PR #343, opened 2022-05-17
Update installer.nsi

### LINKED ISSUE #344: Installer desktop shortcut FIX for all users on pc
I have an admin and user account on my win10 machine and installed vrcx and noticed a desktop shortcut wasnt being created. I tried multiple things(refreshing desktop, toggling the option in the installer) with no luck.

Ended up finding the vrcx shortcut on my admin accounts desktop. A PR is already awaiting approval but wanted to make an issue to help with documentation more.

JayNasty:
Fixed desktop shortcut icon installer by changing the installer desktop function to only apply on the current active users desktop. Bug was found while installing VRCX on my user account and noticing the shortcut only going to my admins desktop.
fixes #344

### DIFF
--- a/Installer/installer.nsi
+++ b/Installer/installer.nsi
@@ -106,7 +106,7 @@ Function .onInit
 FunctionEnd
 
 Function createDesktopShortcut
-    CreateShortcut "$DESKTOP\VRCX.lnk" "$INSTDIR\VRCX.exe"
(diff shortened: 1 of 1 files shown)


Natsumi-sama:
https://nsis.sourceforge.io/Reference/SetShellVarContext
`If set to 'all', the 'all users' shell folder is used.`
Creating a shortcut on every Windows user account desktop isn't a desirable outcome and generally isn't what installers do by default.

PR #388, opened 2022-08-05
Update README.md concerning VRChatRPC.dll

OptoCloud:
I wonder if its better to just remove the entire section completely


SalbugVR:
Would perhaps suggest it to be more like this. Still relatively amateur to Github so don't know how to directly add suggestions into the commit.

image

OptoCloud:
Yeah that works, but I was thinking about keeping the README clean 🤔 
If people keep leaving stuff in and adding more notices to it it quickly becomes very cluttered

PR #556, opened 2023-06-02
Add /settings endpoint, clean up README, make WorldDB U# impl easier/faster

GroovyTeacup:
- Made the README look prettier.
- Added a /settings endpoint for setting allow_external_reads for worlds to true. No way to use this column yet.
- Worlds are now initialized automatically and the connectionKey is now returned with every data request, making the /init endpoint un-necessary in most cases. This allows near-instant data retrieval instead of waiting 5 seconds in-between /init and for example, /get because of the 1 request/5s limitation on String Loading(lmao, btw)
- Also fixed store request logging being completely wrong and broken

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,270 +1,130 @@
-# VRCX
+<div align="center">
 
-[![GitHub Workflow Status](https://github.com/vrcx-team/VRCX/actions/workflows/github_actions.yml/badge.svg)](https://github.com/vrcx-team/VRCX/actions/workflows/github_actions.yml)
+# <img src="https://raw.githubusercontent.com/vrcx-team/VRCX/master/VRCX.ico" width="64" height="64"> </img> VRCX 
+[![GitHub release](https://img.shields.io/github/release/vrcx-team/VRCX.svg)](https://github.com/vrcx-team/VRCX/releases/latest)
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -1,270 +1,130 @@
...
+  - **Note**: To use this feature, you must have "Allow Untrusted URLs" enabled in your VRChat settings.
+  - For Developers: [Wiki Page - World Persistence (PWI)](https://github.com/vrcx-team/VRCX/wiki/World-Persistence-(PWI))
+- :mag: Search for avatars, users, worlds, and groups
+- :clipboard: Build a local, unrestricted world favorites list
+- :camera: Store world data in the pictures you take in-game, so you can remember that one world you took those cool pictures in like... 6 months ago!
+- :bell: Monitor/respond to notifications
+  - You can send/receive invites and friend requests from VRCX as well as see the instance info of invites that you receive.
+- :scroll: See stats/players for your current instance
+- :tv: See the links to videos and that are playing in the world you're in, as well as various other logged data.
+- :bar_chart: Better Discord Rich Presence
```
TayouVR, line 41:
```suggestion
- :bar_chart: Improved Discord Rich Presence
```
`Better Discord` could lead to people thinking this would have something to do with [Better Discord](https://betterdiscord.app/)

PR #563, opened 2023-06-05
update jp translation

Assault1892:
i apologize for the delay, i have been busy with schoolwork and haven't been able to get into the translation


Natsumi-sama:
No rush at all, thanks for spending the time to create translations.

Issue #577, opened 2023-06-16
[Bug] Auto Launch not closing apps after closing VRchat


JanieUwU:
**Describe the bug**
Auto launch setting to auto close does not shut down the program.

**To Reproduce**
1. Put app shortcut into startup folder.
2. Launch VRC, app auto launches with game.
3. Close VRC, program stays open.

**Expected behavior**
Upon closing the game, the apps that was launched with it would also be closed.

**What version you are running**
Version VRCX 2023.06.15

Natsumi-sama:
is it some apps or all apps? does it run as admin?

JanieUwU:
> 

Seems to just be one app. I was using it for a custom made program for moderation of certain in game communities so it's not signed at all, the app should not be running in admin either though.

JanieUwU:
Not sure if it matters but it was built with Electron, using electon-builder's Squirrel.Windows packer?

JanieUwU:
Just checked the app, it is not running as admin

JanieUwU:
How does VRCX know what window to close? It could be that our app's process has different names listed for it's running processes? The shortened processes are just `Moderator Toolkit` while the window's name is `{Community Name] Moderator Toolkit`. Maybe that could be causing the issue?

GroovyTeacup:
I just tested it with a random electron app and it seemed to work, so I'm not sure if electron is an issue.

When the launcher starts a shortcut, it stores the `Process` instance of whatever got spawned by `Process.Start`. When things need to close, it uses the PID of the process it started to close that process and any child processes. Not the window names or handles directly.

If the app, say, starts up a with one process when launching, starts a bunch of child processes, and then kills the parent process without closing the children, VRCX will lose it because it's assumed the process has closed by itself or the user has closed it.

Do you get a UAC prompt when starting the app? AFAIK, UAC prompts will cause this issue as well, as after you accept the prompt, the original process is closed and then re-opened under another PID, also leading to VRCX losing track of the process.

DorCoMaNdO:
> If the app, say, starts up a with one process when launching, starts a bunch of child processes, and then kills the parent process without closing the children, VRCX will lose it because it's assumed the process has closed by itself or the user has closed it.

[...]

PR #578, opened 2023-06-16
Improve monitoring of auto-started apps' processes.

DorCoMaNdO:
Potentially addresses #577.

This PR does not use WMI as originally suggested and instead reuses `Process32First`/`Process32Next` to find child PIDs.

I've not tested whether this addresses UAC, however I did test shortcuts to batch files with `start "" "path\to\program"` commands to simulate intermediate processes, these programs were correctly terminated with the changes in this PR.

### DIFF
--- a/AutoAppLaunchManager.cs
+++ b/AutoAppLaunchManager.cs
@@ -4,6 +4,7 @@ using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Runtime.InteropServices;
+using System.Timers;
 
 namespace VRCX
 {
@@ -21,7 +22,9 @@ namespace VRCX
         public readonly string AppShortcutDirectory;
 
         private DateTime startTime = DateTime.Now;
-        private Dictionary<string, Process> startedProcesses = new Dictionary<string, Process>();
+        private Dictionary<string, HashSet<int>> startedProcesses = new Dictionary<string, HashSet<int>>();
+        private readonly Timer childUpdateTimer;
+        private int timerTicks = 0;
         private static readonly byte[] shortcutSignatureBytes = { 0x4C, 0x00, 0x00, 0x00 }; // signature for ShellLinkHeader
 
         private const uint TH32CS_SNAPPROCESS = 2;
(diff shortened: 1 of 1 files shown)


JanieUwU:
If it gets released on a nightly, I will test it on the app in question! You are awesome <3

DorCoMaNdO:
> If it gets released on a nightly, I will test it on the app in question! You are awesome <3

You can try the artifact from https://github.com/DorCoMaNdO/VRCX/actions/runs/5299550541 (extract and run).

JanieUwU:
> > If it gets released on a nightly, I will test it on the app in question! You are awesome <3
> 
> You can try the artifact from https://github.com/DorCoMaNdO/VRCX/actions/runs/5299550541 (extract and run).

It works! App now gets caught and shut down! Thank you so much <3

Natsumi-sama:
@DorCoMaNdO Would you be willing to take a look at adding something to keep track of running AppX packages as well? for example [VRCFaceTracking](https://github.com/benaclejames/VRCFaceTracking)

DorCoMaNdO:
> @DorCoMaNdO Would you be willing to take a look at adding something to keep track of running AppX packages as well? for example [VRCFaceTracking](https://github.com/benaclejames/VRCFaceTracking)

I've only just seen this, apologies.

I've not interacted with AppX installations (outside of the Windows Store) more than a handful of times, and never in code, so I'm not sure I'm the best choice to tackle this, I did however dig around a little, found some samples [here](https://learn.microsoft.com/en-us/windows/win32/appxpkg/package-deployment-api), the `FindPackages` and `FindPackagesByNameAndPublisher` samples seem helpful to find which applications exist, and the [Package](https://learn.microsoft.com/en-us/uwp/api/windows.applicationmodel.package?view=winrt-22621) class could help by providing the [InstalledPath](https://learn.microsoft.com/en-us/uwp/api/windows.applicationmodel.package.installedpath?view=winrt-22621), which could be used to monitor/terminate apps by checking a process's full path and seeing if it's within the package path, this may be costly but it would only need to run after VRChat runs and/or on close, since it'd only need to match the path to a process, which can then (probably?) be monitored/terminated like any other process.

If this method is acceptable (or you're aware of a better one) then I could test a proof of concept and draft a PR, I could use help on the UI/JS front if a package list does need to be generated, since I'm not too familiar with that part of the codebase, I've only dug around some C# bits.

PR #593, opened 2023-07-03
update jp translation to nightly 2023-07-01T20.53

Assault1892:
The text on lines 801-803 has been withheld because we could not find it in the application.
If the text is only visible to group owners, etc., please let us know via screenshot!


Natsumi-sama:
Ah the lines you are referring to are used in the group dialog members tab when you have moderation permissions for the group you can see this extra info as icons with tooltips. 
Screenshot 2023-07-15 042425

PR #598, opened 2023-07-14
Open world cache directory

BoatFloater:
Sometimes I need to access the cache directory of my worlds, so it would be useful to have this button.

This implementation is pretty quick and dirty, so if you would accept this feature, please let me know any changes you'd like to see.


Natsumi-sama:
I've reworked this but it should still work the same.

PR #666, opened 2023-10-10
Translation of the posts section into Russian

Sergey004:


Natsumi-sama:
Posts are group announcements

PR #673, opened 2023-10-20
fix: Verify monitored process exit state before adding & emitting event

GroovyTeacup:
This (probably) fixes the bug where a monitored process would be constantly set as running/not running because it would be constantly removed and then re-added every second.

Also added some logging to the process monitor


Natsumi-sama:
Merge conflict, deadge

PR #693, opened 2023-11-22
Added self contained parameter to builds

DubyaDude:


DubyaDude:
@Myrkie @Natsumi-sama @BenjaminZehowlt
Mind giving this a test run to see if anything significant breaks?

DubyaDude:
@Okabintaro Can you try this install?
https://github.com/vrcx-team/VRCX/suites/18400997359/artifacts/1066185171

Okabintaro:
> @Okabintaro Can you try this install? https://github.com/vrcx-team/VRCX/suites/18400997359/artifacts/1066185171

That one works fine! I also made an issue for my previous problem in #695.

TayouVR:
could this be an extra build artifact, offering two versions; one self contained and one using the system .net install?

DubyaDude:
> could this be an extra build artifact, offering two versions; one self contained and one using the system .net install? Though I'm not sure how that would work with the updater, I guess it would have a flag or smth to know to update to the self contained or the other version.

While that's something that could be done, would require effort.
I don't see the cons listed as being enough to push us towards that. If you're short on disk space to be worried by 160MB, I think you will run into storage issues with or without self-contained.
And regarding the 'we control runtime version', it'll be upgraded to the latest every build by the GitHub runner, so keeping it updated isn't really tedious.

I don't see any significant enough of a downside of this.

Natsumi-sama:
Sometimes more options isn't better, this certainly one of those times, it's not worth the time explaining the difference in builds every time to people that just don't care or notice.

TayouVR:
> it's not worth the time explaining the difference in builds every time

fair, thought of that too, I just like having options. (and I like having my dependencies installed system wide - shared across apps)

Natsumi-sama:
> Mind giving this a test run to see if anything significant breaks?

It seems building with `--self-contained` causes `CefSharp.BrowserSubprocess.exe` to be excluded, we could go down the route of building our own `BrowserSubprocess` into the main exe I've seen that done by other people using Cef.

DubyaDude:
> > Mind giving this a test run to see if anything significant breaks?
> 
> It seems building with `--self-contained` causes `CefSharp.BrowserSubprocess.exe` to be excluded, we could go down the route of building our own `BrowserSubprocess` into the main exe I've seen that done by other people using Cef.

I removed that from the csproj though, should be included now, no?
I tested a quick launch using Sandbox, dotnet installed, using the build artifact here and it seems to work fine at login, haven't tested past that.
```diff
- <CefSharpExcludeSubProcessExe>true</CefSharpExcludeSubProcessExe>
```

Natsumi-sama:
Yeeeeea that was it, I'm not sure how I manged to miss that.

PR #768, opened 2024-04-12
Update flags to support the newly added languages

hyblocker:
This PR updates the flags so that [the recently added languages (to user bios)](https://ask.vrchat.com/t/developer-update-11-april-2024/23928) have their appropriate flags.

yue (Yue Chinese) and tws (Teochew Min) evaluate to `??` resulting in the unknown flag, as I am unsure on which specific flag should be used for these.


Natsumi-sama:
Ah! thanks, I was dreading having to deal with this. I have no idea what to do about the missing flags I'll ask some people that might know.

hyblocker:
I've mapped them to the Chinese flag now then.

PR #849, opened 2024-07-25
Enhancement for the Linux Install Script

guerryer:
I made it so it always grab the latest Stable or Nightly releases in the script with grep.

### DIFF
--- a/Linux/install-vrcx.sh
+++ b/Linux/install-vrcx.sh
@@ -3,7 +3,9 @@
 
 # change me
 steamapps=$HOME/.local/share/Steam/steamapps/compatdata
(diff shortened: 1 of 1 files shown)


regalialong:
Ah, this is sort of intentional, both because I wanted to avoid parsing from JSON without an actual JSON parser and because I don't trust a CEF version bump to not break Wine.

But realistically, unless we get some sort of option to wrangle the Autoupdater (especially before a database actually exists), this is moot so we might as well follow latest.

regalialong (approved):

MaximeMichaud (approved):
LGTM

PR #864, opened 2024-08-06
feat: kill install process after exiting game

lonelyicer:
should fix https://vrchat.canny.io/bug-reports/p/installexe-breaks-osc-port-binding

### DIFF
--- a/Dotnet/AppApi/GameHandler.cs
+++ b/Dotnet/AppApi/GameHandler.cs
@@ -15,7 +15,7 @@ namespace VRCX
 
             CheckGameRunning();
         }
-        
+
         /// <summary>
         /// Checks if the VRChat game and SteamVR are currently running and updates the browser's JavaScript function $app.updateIsGameRunning with the results.
         /// </summary>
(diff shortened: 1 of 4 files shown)


Dotnet/AppApi/GameHandler.cs
```diff
@@ -54,6 +54,19 @@ public int QuitGame()
             return processes.Length;
         }
 
+        /// <summary>
+        /// Kills the install.exe process after exiting game.
+        /// </summary>
+        /// <returns>The number of processes that were killed (0 or 1).</returns>
+        public int KillInstall()
+        {
+            var processes = Process.GetProcessesByName("install");
```
DubyaDude, line 63: I'm a bit hesitant about the way `install.exe` is being grabbed here.  While it makes sense for `vrchat` because the likelihood of the same application existing is little to none, `install` can very much encompass more processes that may be unintentionally killed.

DubyaDude (changes requested):

Dotnet/AppApi/GameHandler.cs
lonelyicer: Detect if install.exe is in VRChat folder with regular expression

PR #868, opened 2024-08-11
fix: sometimes install.exe is suspended

lonelyicer:


Dotnet/AppApi/GameHandler.cs
```diff
@@ -64,6 +66,9 @@ public bool KillInstall()
             var processes = Process.GetProcessesByName("install");
             foreach (var p in processes)
             {
+                // Sometimes install.exe is suspended
+                ResumeProcess(p.Id);
```
DubyaDude, line 70: This would need to be in the `if (match.Success)` part of the code.
Again, to make sure we're not messing up any other processes by accident

DubyaDude (changes requested):

Dotnet/AppApi/GameHandler.cs
DubyaDude: @lonelyicer 
```
// Sometimes install.exe is suspended
ResumeProcess(p.Id);
```

should be moved inside
```
if (match.Success)
```

so we don't affect other unrelated install processes

lonelyicer:
@DubyaDude Use WinBase.QueryFullProcessImageName instead of Process.MainModule.FileName.

Issue #875, opened 2024-08-18
VRCX API Update Prep


DubyaDude:
This is an issue to track preparation of the API upgrade. Currently the API is live for testing under https://api1.vrcx.app/
Source: https://github.com/vrcx-team/VRCX-API

## The main goals for this API change is:
- Parity with the current API to make this a drop in replacement with V0 once tested.
- Allow us to utilize Cloudflare caching to reduce loads on the server.
- Allow us to merge nightly (Natsumis fork) and main repository (vrcx-team) into one repository.
- Allow for easy future expansion.

## How does the API differ from current?
- You may see some fields be swapped out/new fields as the data model isn't 1:1, but for the purpose of the VRCX updater, the fields that matter should be identical.
- `stable` will contain all releases from the main repository that are not marked as a pre-release.
- `nightly` will contain all releases from the main repository that are marked as a pre-release merged with all releases from the nightly fork.
- The `prerelease` field will always be false due to some legacy logic.

## Checklist
- [X] Release API for testing under v1 subdomain
- [x] [Remove legacy updater logic](https://github.com/vrcx-team/VRCX/blob/a7cad548672aabf306b5b89de233fc1b4a06749e/html/src/app.js#L25723-L25729).
  - https://github.com/vrcx-team/VRCX/pull/876
- [x] Remove all references of grabbing GitHub releases directly in favor for `https://api0.vrcx.app/`
  - https://github.com/vrcx-team/VRCX/pull/876
- [x] Change the old pypys domain to redirect to `https://api0.vrcx.app/`.
- [x] Deploy new API under v0 subdomain.

DubyaDude:
API is now live under https://api0.vrcx.app/

PR #876, opened 2024-08-18
VRCX API Update Changes

DubyaDude:
- Removed legacy logic from the updater
- Updated Linux install script to utilize api

### DIFF
--- a/Linux/install-vrcx.sh
+++ b/Linux/install-vrcx.sh
@@ -1,8 +1,8 @@
 #!/usr/bin/env bash
 
 steamapps=$HOME/.local/share/Steam/steamapps/compatdata
-stable=$(curl -s https://api.github.com/repos/vrcx-team/VRCX/releases/latest | grep -o "https.*releases/download.*zip")
-nightly=$(curl -s https://api.github.com/repos/Natsumi-sama/VRCX/releases/latest | grep -o "https.*releases/download.*zip")
+stable=$(curl -s https://api0.vrcx.app/releases/stable/latest | grep -o "https.*releases/download.*zip")
+nightly=$(curl -s https://api0.vrcx.app/releases/nightly/latest | grep -o "https.*releases/download.*zip")
 download_url=$stable
 XDG_DATA_HOME=${XDG_DATA_HOME:=$HOME/.local/share}
 
--- a/html/src/app.js
+++ b/html/src/app.js
@@ -25720,13 +25720,6 @@ speechSynthesis.getVoices();
             return;
         }
         for (var release of json) {
-            if (
-                release.target_commitish === 'PyPyDanceCompanion' ||
-                release.prerelease
-            ) {
-                // skip old branch name and prerelease builds
-                continue;
-            }
             for (var asset of release.assets) {
                 if (
                     (asset.content_type === 'application/x-msdownload' ||


Linux/install-vrcx.sh
```diff
@@ -1,8 +1,8 @@
 #!/usr/bin/env bash
 
 steamapps=$HOME/.local/share/Steam/steamapps/compatdata
-stable=$(curl -s https://api.github.com/repos/vrcx-team/VRCX/releases/latest | grep -o "https.*releases/download.*zip")
-nightly=$(curl -s https://api.github.com/repos/Natsumi-sama/VRCX/releases/latest | grep -o "https.*releases/download.*zip")
+stable=$(curl -s https://api0.vrcx.app/releases/stable/latest | grep -o "https.*releases/download.*zip")
```
regalialong, line 4: What's this change for? `api0.vrcx.app` seemingly just proxies the release endpoint. The bash script has no cryptographic / hash verification for the binaries themselves, I'm kinda queasy about moving this to a server without published source code.

DubyaDude: https://github.com/vrcx-team/VRCX/pull/876

DubyaDude: let me just re-comment since I linked the wrong thing.
https://github.com/vrcx-team/VRCX/issues/875
API0's content will soon change.

DubyaDude: We plan to merge nightly and main repo, which will break this.
As stated in the issue above:
- `stable` will contain all releases from the main repository that are not marked as a pre-release.
- `nightly` will contain all releases from the main repository that are marked as a pre-release merged with all releases from the nightly fork.

DubyaDude: If the source of the API app is the issue, I'll talk with Natsumi about how we want to go forward with that. I'm not opposed to open-sourcing it.

regalialong: Why would that change require a server? The `latest` endpoint according to docs will always return a non pre-release, meanwhile releases itself contains the `prerelease` field.

Of course to do this here it'd require a JSON parser and some logic, which isn't necessarily a problem. But I doubt the API was introduced for the convenience of the Linux script ^^'

DubyaDude: If you want to mimic the API in the Linux script itself, sure be my guest. This was primarily for parity with the VRCX app updater itself which is already calling api0, along with the legacy domain now pointing towards it.

I'd prefer it to have the vrcx.app server call for the sake of consistency, so if there's anything you'd want that would make you comfortable with it, I'd be more than willing to make amends.

regalialong: It's more or less that this is swapping to a third party server for seemingly no tangible gain on the Linux side.

An obvious convenience advantage here would be if you could pass the release URL directly instead of us needing to parse it here per grep.

On the paranoid side, cheapest thing would obviously be to have the API be auditable by open sourcing the server side.

A way to fix the API trust issue would be to pull the SHA256 from Github and check that, however _SHA256SUMS.txt_ only contains the one for the `VRCX_*_Setup.exe` file :^)

DubyaDude: Not opposed to open-sourcing, just want to discuss with Natsumi how we want to handle repository ownership then once I clean up the code a little.

The main reason for the switch is the plan for a single repository rather than having a fork for nightlies.
If GitHub had a proper way to filter by release/pre-release, this wouldn't have been an issue.

Do be aware of the following though when you try to mimic the API itself:
- Will need to do pagination, as there wouldn't be a guarantee of the prerelease existing on the first page.
- Also the transition to a single repo won't be immediate, so you'd need to juggle both repos for the latest nightly.
- Github rate limits for unauthenticated requests are about 60/hr, which you could hit often.

DubyaDude: You could theoretically just get the release ID from the API, query the GitHub API for the specific release, and grab the DL from that if open-sourcing isn't enough.

DubyaDude: And, to address the convenience factor of the Linux script, I don't mind expanding the API to make it easier for this script to function. One of the goals of this 'update' was parity with the previous API to not break older clients, so as long as that's still maintained, I'm not opposed to new endpoints.

regalialong: The releases endpoint seems to be sorted by the latest releases, and allows returning a maximum of 100 releases. 100 releases without a pre-release seems like a weird scenario.

Github ratelimits are a problem, yeah, but there's not much we can do about that aside fail. I don't think unattendedness was an explicit goal of the script.

> Also the transition to a single repo won't be immediate, so you'd need to juggle both repos for the latest nightly.

I don't have anything nice to say about this.

Anyway yeah, if those two are done, I think the change would be fine. Obviously my authority here is phony but that's just my opinions on it.

DubyaDude: Source: https://github.com/DubyaDude/VRCX-API
I'll likely have it moved under the `vrcx-team` GitHub org soon.
The only change I had planned from this is to do proper configuration as it's currently hooking into a configuration file used by some of the other apps on the server.

DubyaDude: It's been moved to https://github.com/vrcx-team/VRCX-API
I've made some of the discussion points into issues, feel free to tack on things if interested: https://github.com/vrcx-team/VRCX-API/issues

PR #912, opened 2024-09-11
Linux: Use the VRCX.app direct links intead

regalialong:
I suppose I should stick to my word, I had the grep workaround fail on my end so it's more than overtime to just use the direct URLs instead.

### DIFF
--- a/Linux/install-vrcx.sh
+++ b/Linux/install-vrcx.sh
@@ -1,8 +1,8 @@
 #!/usr/bin/env bash
 
 steamapps=$HOME/.local/share/Steam/steamapps/compatdata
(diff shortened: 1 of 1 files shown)


DubyaDude:
@Natsumi-sama any harm in expediting this one since it's just fixing the linux install script?

DubyaDude:
We cannot pull this at the current moment.
For the time being, Please run the following to run this fixed version of the install script:
```
curl -sSf https://raw.githubusercontent.com/vrcx-team/VRCX/4d63828f44fc29388cd874264cc8462247e1d874/Linux/install-vrcx.sh | bash
```

PR #942, opened 2024-10-18
Linux: Fix .desktop to match WM_CLASS

rs189:
Renamed the name of the .desktop file from `vrcx.desktop` to `vrcx.exe.desktop` as the WM_CLASS of VRCX used is `WM_CLASS(STRING) = "vrcx.exe", "vrcx.exe"` as per xprop.

This means that now instead of the VRCX showing up as 
image

It now shows up as
image

Only tested on Fedora 40 on Wayland and X11 for now


regalialong (approved):
I didn't notice with this on KDE Wayland, but I believe this is the same name that KDE matches against and I didn't notice anything breaking from this, so greenlight from me if this fixes the icon on GNOME. Thanks.

PR #946, opened 2024-10-19
Rework Startup Args

DubyaDude:
- Rework startup arguments handling.
- Prevent duplicate instances of the same config directory from starting.
- Preserve startup arguments on upgrade restart.


DubyaDude:
Need to do before change from Draft to PR
- [ ] Get the difference between subprocess/main process when checking the duplicated config directory (To support having an instance launched with a custom directory, then launching an instance with no custom directory)
- [ ] Proper testing, both in Windows and Linux. (I don't expect the added duplicated config directory to work in Linux, just want to make sure it doesn't fail catastrophically)

DubyaDude:
Ok, I'm actually done now, should be good to go

PR #959, opened 2024-10-25
Request High Performance GPU

BenjaminZehowlt:
Microsoft does not provide a native mechanism to do this any other way currently in .NET


Myrkie:
This is awful but so on brand for Microsoft.

PR #991, opened 2024-11-24
Linux: Use NSIS installer

regalialong:
Okay, this PR is mostly minor changes with bigger implications.

I implemented the `SKIP_SHORTCUT` option for the Setup:
- When Wine notices that a shortcut has been installed, it helpfully creates a `.desktop` file for us so that shortcuts are accessible from Linux.
- We provide our own `.desktop` file (partially because I don't think DEs like the autogenerated one?), so this isn't very helpful to us, hence why you can now pass `/SKIP_SHORTCUT=true` to the silent installer and it'll skip making the shortcuts.

I also changed the install script to now use the setup instead of unzipping ourselves, cutting out the `vrcx` script in the process since it's mostly a remnant of when vrcx was still getting added to PATH.

Next up, I added Wine detection in dotnet by calling `wine_get_version` per a `ntdll` extern, that should only exist through Wine.

Tying this all together, the Updater passes `DISABLE_SHORTCUT` if we're on Wine.
 
The rationale for this is that Linux users can use the auto updater now. Not having any way to tell the updater to stop is a big pain point since it can create a second shortcut to VRCX in another install location, that points directly to the executable instead of our script. 

I think this is a very weird UX and I honestly didn't have a good answer on how well that works up until now.

I'm pretty sure that making a special mode for this would be more pain than it's worth, so this also aligns Linux environments more with Windows.

### DIFF
--- a/Dotnet/Program.cs
+++ b/Dotnet/Program.cs
@@ -178,6 +178,7 @@ namespace VRCX
             Application.SetCompatibleTextRenderingDefault(false);
 
             logger.Info("{0} Starting...", Version);
+            logger.Debug("Wine support detection: {0}", Wine.GetIfWine());
             
             ProcessMonitor.Instance.Init();
             SQLiteLegacy.Instance.Init();
--- a/Dotnet/Update.cs
+++ b/Dotnet/Update.cs
@@ -35,6 +35,9 @@ namespace VRCX
 
         private static void Install()
         {
+            var setupArguments = "/S";
+            if (Wine.GetIfWine()) setupArguments += " /DISABLE_SHORTCUT=true";
+            
             try
             {
                 File.Move(Update_Executable, VRCX_Setup_Executable);
@@ -43,7 +46,7 @@ namespace VRCX
                     StartInfo = new ProcessStartInfo
                     {
                         FileName = VRCX_Setup_Executable,
-                        Arguments = "/S",
+                        Arguments = setupArguments,
                         UseShellExecute = true,
                         WorkingDirectory = Program.AppDataDirectory
                     }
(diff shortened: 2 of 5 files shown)


rs189:
I haven't had the time to test this, but would your change break https://github.com/vrcx-team/VRCX/pull/992?

rs189:
Testing this on openSUSE Tumbleweed returns `bash: cd: too many arguments` on line 72, this is due to `$INSTALL_LOCATION` having spaces, which can be easily fixed by adding quotation marks `cd "$INSTALL_LOCATION"`

regalialong:
Oops, I think I spotted that but not sure why I didn't commit this.

regalialong:
If you want to test this properly, you'll have to build the installer from this PR (since this PR also modified the NSIS script) and make sure it pulls that instead from curl since latest installer right now doesn't have the functionality to suppress the shortcut.

The codeless alternative would be making sure that the shortcut for `~/.local/share/applications/wine/Programs` can't be touched but I think that's a hack :^)

rs189:
I could try that, just so I understand this right, the NSIS installer in this implementation is completely headless and is not supposed to have any GUI on Linux, right?

regalialong:
Yeah, NSIS allows unattended silent installation by passing `/S`, I think it gets that as part of NSIS.

rs189:
In that case, after this is merged, perhaps we could make it into a Flatpak?

regalialong:
Probably worth discussing over in Discord but think this was discussed in an internal chat, I remember Ben thinking that CEF probably won't work in bwrap.

That and we need access to the filesystem for game log, I'm sure it's possible to punch a hole through Flatpak but I'm not sure if it's worth it in the long run besides for bundling a Wine version.

PR #992, opened 2024-11-24
Linux: Fix emojis not showing in the game log

rs189:
VRCX on Windows relies on Segoe UI Emoji font to render emojis in the game log to indicate a friend or a favourite, as this font is proprietary it cannot be used on Linux, I have remapped Segoe UI Emoji to use [twemoji](https://github.com/twitter/twemoji) font instead (MIT licence)

It lacks colour but properly shows up

image

Tested on Fedora 41 and OpenSUSE Tumbleweed

### DIFF
--- a/Linux/install-vrcx.sh
+++ b/Linux/install-vrcx.sh
@@ -78,6 +78,15 @@ export WINEPREFIX=$WINEPREFIX
 wine $WINEPREFIX/drive_c/vrcx/VRCX.exe" > $WINEPREFIX/drive_c/vrcx/vrcx
 chmod +x $WINEPREFIX/drive_c/vrcx/vrcx
 
+rel_path=$(dirname $(realpath $0))
+
+# Install twemoji fonts as Segoe UI is proprietary and not included in wine
+if [[ -d "$rel_path/fonts" ]]; then
+    echo "Installing twemoji fonts."
+    cp -r "$rel_path/fonts/"* "$WINEPREFIX/drive_c/windows/Fonts"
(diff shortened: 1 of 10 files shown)


Linux/install-vrcx.sh
```diff
@@ -78,6 +78,15 @@ export WINEPREFIX=$WINEPREFIX
...
 chmod +x $WINEPREFIX/drive_c/vrcx/vrcx
 
+rel_path=$(dirname $(realpath $0))
+
+# Install twemoji fonts as Segoe UI is proprietary and not included in wine
+if [[ -d "$rel_path/fonts" ]]; then
+    echo "Installing twemoji fonts."
+    cp -r "$rel_path/fonts/"* "$WINEPREFIX/drive_c/windows/Fonts"
+	WINEPREFIX=$WINEPREFIX wine reg add 'HKLM\Software\Microsoft\Windows NT\CurrentVersion\Fonts' /v 'seguiemj' /t REG_SZ /d 'seguiemj.ttf' /f
+fi
```
regalialong, line 88: If I understand this correctly, doesn't this depend on a user pulling the VRCX repo? I think most users only execute the script directly as per curl, think that'd never pull the fonts folder.

rs189: Good catch, I will update the script to curl the font file from the repo.

regalialong (approved):

rs189:
I have completely replaced the system emoji font usage with [Noto Color Emoji](https://fonts.google.com/noto/specimen/Noto+Color+Emoji?query=emoji) via [@infolektuell/noto-color-emoji](https://github.com/infolektuell/noto-color-emoji) npm package, this change will affect everyone not just the Linux version; I have kept the earlier change adding [twemoji](https://github.com/13rac1/twemoji-color-font) as a system emoji font for Wine for further compatibility. I believe this should be good to merge now.

PR #1018, opened 2024-12-09
i18n: add Chinese and Japanese language support for message box

Map1en:
feat: add Chinese and Japanese language support for message box

- Add translations for dialog content in zh-CN and ja-JP
- Previously English only


Natsumi-sama:
Oh this is great to see, this has been on mine and few other peoples todo list for a long time.

PR #1022, opened 2024-12-11
i18n: add zh-cn and ja-jp translations for API error messages

Map1en:
- Add zh-CN and ja-JP translations for HTTP error messages

> Although HTTP status codes in English are universal, for regular users, having error messages in their local language makes it easier to understand and troubleshoot issues.

- Remove redundant escape characters and quotes in error messages


Natsumi-sama:
Status codes seems a bit excessive, but no harm I guess so why not.

PR #1034, opened 2024-12-16
UI: Enhancement (#1033)

Map1en:
- issue #1033

----------------------------------------------------------------
> 0570716

Fixed the following issues:
1. Main Item: User Dialog Box
   - Adjust the height of the text in the tags to be centered.
   - Change the background color of the counter when hovering over VRC notes.
   - Re-align the radiogroup and refresh button for model creation, and also re-align the top of the created worlds and groups.

2. Avatar Dialog Box
   - Leave a space below the local notes.

3. Main Screen
   - Align the input boxes on each page with the sidebar friend list.
   - Avoid using both `v-if` and `v-show` simultaneously; remove conflicts.

4. Search Avatar Page
   - Re-align the buttons and radio at the top.

5. Settings
   - Re-align and adjust the VRC configuration page.
   - Re-align and adjust the data backup page.
   - Adjust the spacing of the notification filters.

--------------------------------------------------
> cdf5688193baa29c31547541c407aa7bb9b6a1d3
cb22cd0c24e52e66c971ba3726f5896c2ea30f14

Fixed the following issues:
1. Background color issue when hovering over input count in non-default themes
2. Adjusted the top height of the feed page
3. Fixed margin issues for detailed information when status changes in the feed page

--------------------------------------------------
> 3cce05c

Fixed the following issues:
1. Feed Page
   - Fixed spacing between status items

2. Favorites Page
   - Search and Sort UI
     • Reorganized search box and sorting methods to top-right corner
     • Replaced Switch with RadioGroup for sorting methods (as they represent distinct values, not binary states)
     • Similar components will be updated in future PRs
   - Visual Improvements
     • Restored original colors for friend avatars (removed grey tint, Because it is inconsistent with the behavior of the friends status bar on the right side.)
     • Offline status icons updates pending further consideration
   - Data Display
     • Added "No data" placeholder in empty Collapse components
     • Replaced dot indicators with tags for status of world favorites (It can easily cause confusion, we switched to using tags to improve recognition.)
   - Import/Export Dialog Features
     • Redesigned import dialog layouts across all three tabs
     • Enhanced export dialogs:
       - Added click-to-copy functionality for textarea content
       - Implemented success/failed notifications
   - UI Polish
     • Adjusted text case displays for consistency
     • Basic theme compatibility improvements
       (Note: Comprehensive theme adjustments pending due to color hierarchy complexity)

----------------------------------------------------------------------------------------------------
> 20964f7

_I'm sorry for making such fragmented commits and not following the specific pages._

[...]

Issue #1033, opened 2024-12-16
[UI] Enhancement


Map1en:
This is a tracking issue for UI improvements. I've been working on some UI enhancements recently, I think it's a good idea to create an issue to track these changes.

This includes basic alignment issues, layout organization, and some UX adjustments afterwards.

- Alignment issues
- Layout organization
- UX adjustments

Currently, modifications are made according to the default theme. Later, additional changes will be made to other themes based on the existing style.

If possible, I would appreciate some suggestions so I can have a clear direction for improvement, Thanks.

### Related PRs:
- #1034 
- #1047 
- #1054 
- #1062
- #1078
- #1103 
- #1122

Natsumi-sama:
Kamiyas Material 3 theme supports https://www.npmjs.com/package/@material/material-color-utilities but I never got around to adding it.
They also made another theme that was never added https://github.com/kamiya10/VRCX-theme/tree/main/bg
Maintenance cost of large amounts of themes can add up.
Our UI library element-ui everything is based off of is deprecated.
Upgrading would require switching to Vue3 or another JS framework, essentially requiring a full rewrite making it not a viable option.
These fixes and improvements are most welcome.

Map1en:
First of all, the current program's functionality is excellent. However, there are many aspects of the UI that could be improved.

Under the default theme, there are already several issues as described in the posts above. My plan is to start with alignment and arrangement: ensuring that most elements are properly aligned and displayed in a more intuitive way.

Next step would be unifying UI components: currently, many similar interactions are not using the same type of component, which likely needs to be corrected.
Lastly, for better UX - this is a long-term consideration and won't be addressed immediately.

Currently, non-default themes have numerous display issues, far more than those in the default theme. These aren't just basic alignment problems - they appear unattractive and are even difficult to use. Furthermore, when modifying the default theme code and changing the existing DOM structure, adapting other themes requires a considerable amount of work. Based on this, I suggest giving lower priority to adapting other themes. To avoid excessive workload, we might even consider temporarily deprecating some themes or pausing their updates.

PR #1047, opened 2024-12-24
fix: issues caused by PR(#1034). (#1033)

Map1en:
fixed some issues caused by #1034 

1. Correct the reversed setting for showTooltips.

2. Fix the logout button color (not adapted for other themes).

3. Address the missing background color for textarea in the dark theme.

4. Adjust tag text alignment in the MUI3 theme to ensure it is centered.

### DIFF
--- a/Dotnet/AppApi/AppApi.cs
+++ b/Dotnet/AppApi/AppApi.cs
@@ -141,33 +141,49 @@ namespace VRCX
 
         public byte[] ResizePrintImage(byte[] imageData)
         {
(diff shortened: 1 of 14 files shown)


Natsumi-sama:
Quite a few settings were reversed, I'm sure the switch to using `simpleSwitch` mixin is at fault, toggles were inverted using `:active-value="false" :inactive-value="true"`, personally I'm not a fan of using the mixin because it breaks VS Code plugin i18n-ally but I was too lazy to revert existing settings, but as you said before these sorts of things would make much more sense as components but VRCX currently makes very poor use of components.

PR #1056, opened 2024-12-27
Adds a regenerate impostor button

poprox24:
This feature adds a button to regenerate an impostor, to save a few clicks
It deletes the impostor, refreshes the avatar dialog and queues the creation for a new one
Screenshot 2024-12-27 151243


poprox24:
Made a mistake, will fix and open again

PR #1057, opened 2024-12-27
Add a regenerate impostor button

poprox24:
This feature adds a button to regenerate an impostor, to save a few clicks
It deletes the impostor, queues a new one for creation and refreshes the avatar dialog
Screenshot 2024-12-27 153522


poprox24:
I do not understand Asian languages, I would like to request that someone makes a proper translation of the "Regenerate impostor" button as I cannot confirm the validity of the translation I would make

PR #1076, opened 2025-01-11
use libraryfolders.vdf to resolve vrchat steam path on linux

lun-4:
discussed on VRCX discord: https://ptb.discord.com/channels/854071236363550763/1327441458206347294/1327716411568558252


parkeerwacht:
Tested works with no issues on my end so far. My VRC is saved under a different path than the default. I got a separate game drive that auto mounts under /mnt/Games/ Now VRCX autodetects the path and been able to use it.

Donad678:
I have VRC is on a seperate drive mounted under $HOME/mountedDrives/Gamin, detection of the new path works without issues so far with this patch

PR #1086, opened 2025-01-20
Add Avatar Wear Timers

### LINKED ISSUE #635: [Feature Request] Avatar Wear Times
**Explain in detail what your suggested feature would be used for.**
In the Avatar list, keep a time for how long you've worn that avatar in total

**Describe how it would look if it requires a UI.**
On whatever avatar page, similar to how the friends list has a time spent timer

**Explain why people would want to use it.**
1. Get rid / move avatars you don't use much
2. See what others see you as more often
3. Stats are just nice

Tamaaraa:
Adds support for tracking avatar wear times
Adds a new column in avatar_history in the database to track time for each individual avatar
Uses a new parameter in currentUser in the API: $previousAvatarSwapTime which holds the time at which you swapped into an avatar

Closes #635


html/src/app.js
```diff
@@ -19482,24 +19496,40 @@ speechSynthesis.getVoices();
...
+                API.currentUser.$previousAvatar !== avatarId
+            ) {
+                this.addAvatarWearTime();
+            }
+            API.currentUser.$previousAvatar = avatarId;
+            API.currentUser.$previousAvatarSwapTime = Date.now();
+
+            database.addAvatarToCache(ref);
+            database.addAvatarToHistory(ref.id);
+
```
Tamaaraa, line 19510: Had to swap around some parts of this code so that your own avatars would be added to history in the database to track the time, but not the array for avatar history that vrcx uses

PR #1089, opened 2025-01-22
Use Wayland when available

rs189:
Enabled the auto ozone platform for Wayland support when available. I did not test fractional scaling, but that may work too.

### DIFF
--- a/src-electron/main.js
+++ b/src-electron/main.js
@@ -9,6 +9,7 @@ const {
     dialog,
     Notification
 } = require('electron');
+const { spawn } = require('child_process');
 const fs = require('fs');
 const https = require('https');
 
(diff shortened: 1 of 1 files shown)


Donad678:
Just to have it mentioned, resizing the window when running with wayland does still not work properly (neither in vrcx, tried that nor does it in other Electron applications that use an up to date electron version)

Probably related to this upstream electron bug https://github.com/electron/electron/issues/44543

rs189:
Unless I understand the issue incorrectly, I can resize the window just fine. I am using Fedora 41 KDE Spin with the NVIDIA 565 driver.

Donad678:
I'm running Bazzite 41, Gnome, Nvidia 565, the electron issue seems to be primarily affecting gnome, shouldn't be a blocker as long as you still allow people to switch from auto to x11 in case of issues like this (either by allowing an override through parameters or a setting inside the app)

rs189:
Not sure what would be the best way of handling this, perhaps this should be discussed in Discord.

AKoskovich:
> Unless I understand the issue incorrectly, I can resize the window just fine. I am using Fedora 41 KDE Spin with the NVIDIA 565 driver.

The issue only affects GNOME from what I can see.

PR #1096, opened 2025-01-24
feat: favorite friends divide by different favorite group

ryougishiki2333:
feat: favorite friends divide by different favorite group

example


ryougishiki2333:
I have made the update compatible with previous features

PR #1099, opened 2025-01-24
Revert "Czech language for VRCX"

poprox24:
Reverts vrcx-team/VRCX#1071
I just tried to build it with all the changes that happened and it seems there was a change which makes it unable to start with the czech language, I will have to fix that, you can also fix it if you want to

image


Natsumi-sama:
I'll look into it, but looking at that error it might be from something else.

PR #1130, opened 2025-02-07
Linux: Fix AssetBundleManager.GetAssetVersion only working sometimes

regalialong:
On Linux, the cache version directories are lowercase but `versionHex` here is completely uppercase. This would be fine on case insensitive systems but ext4 isn't, hence it will fail the existence check later on in `CheckVRChatCache`.

The annoying part is that if the version hexadecimal does not create letters, the check can work as intended :))))

You don't know how much """"""fun""""""" this was to debug this without a project.

Untested on Windows so please check

### DIFF
--- a/Dotnet/AssetBundleManager.cs
+++ b/Dotnet/AssetBundleManager.cs
@@ -54,7 +54,7 @@ namespace VRCX
                 versionHex += b.ToString("X2");
             }
 
-            return versionHex.PadLeft(32, '0');
+            return versionHex.PadLeft(32, '0').ToLowerInvariant();
         }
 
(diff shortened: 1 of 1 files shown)


Natsumi-sama:
Oh yea these sorts of bugs are the worst to hunt down.

regalialong:
> Oh yea these sorts of bugs are the worst to hunt down.

Oh no, finding it was fine, I figured it was something along the lines.
The actual issue is that the only valid project that Rider takes it the CEF project, the Electron project refuses to load

so I can't do breakpoints in the actual files, I need to attach, halt, and then ask it to disassemble the app dll, then make breakpoints there

PR #1135, opened 2025-02-12
Fix DirectoryNotFoundException in ImageSaving.cs:CropAllPrints

Myrkie:
Fixes a DirectoryNotFoundException when the path "Prints" doesn't exist


Dotnet/AppApi/Common/ImageSaving.cs
```diff
@@ -177,6 +177,11 @@ public byte[] ResizePrintImage(byte[] imageData)
         public async Task CropAllPrints(string ugcFolderPath)
         {
             var folder = Path.Join(GetUGCPhotoLocation(ugcFolderPath), "Prints");
+
+            if (!Directory.Exists(folder))
+            {
+                Directory.CreateDirectory(folder);
+            }
```
Hackebein, line 184:
```suggestion

            if (!Directory.Exists(folder))
            {
                return;
            }
```

PR #1226, opened 2025-05-04
fix(dbmerger): merge reserved table, Exception due null time in avatar_history

book000:
Thank you for developing this application.
This pull request contains changes to DBMerger.

Fix the following two problems:

1. Avoiding errors by merging reserved SQLite tables “sqlite_sequence”
2. Avoid NullPointerException due to newAvatarTime being null

## 1. Avoiding errors by merging reserved SQLite tables “sqlite_sequence”

When performing a merge, it tries to merge the sqlite_sequence table and therefore tries to create a reserved SQLite table (prefixed “sqlite_”), resulting in the following error

```
SQLite.SQLiteException: object name reserved for internal use: sqlite_sequence
```

Exclude tables beginning with "sqlite_" from the merge tables.
Also, to avoid appearing in “Found unmerged table”, skip it after removing it from `unMergedTables`.

## 2. Avoid NullPointerException due to newAvatarTime being null

When merging the "avatar_history" table, there seems to be a case where the time column in the new DB is NULL.

```
System.NullReferenceException: 'Object reference not set to an instance of an object.'
```

In this case, a NullPointerException will be raised and an error will occur, so the case of NULL will be handled as time=0 to avoid the error.

### DIFF
--- a/DBMerger/Merger.cs
+++ b/DBMerger/Merger.cs
@@ -306,7 +306,7 @@ namespace DBMerger
                         return oldDateTime > newDateTime ? old : existing;
                     }
 
-                    var oldAvatarTime = old.Length >= 3 ? (int)old[2] : 0;
+                    var oldAvatarTime = old.Length >= 3 && old[2] != null ? (int)old[2] : 0;
                     var newAvatarTime = existing[2] != null ? (int)existing[2] : 0;
 
                     if (oldDateTime <= newDateTime)
--- a/DBMerger/Merger.cs
+++ b/DBMerger/Merger.cs
@@ -258,8 +258,8 @@ namespace DBMerger
         private void MergeUsers()
         {
             MergeTable(
-                table => userIDRegex.IsMatch(table)
-                    && !table.EndsWith("_avatar_history")
+                table => userIDRegex.IsMatch(table) 
+                    && !table.EndsWith("_avatar_history") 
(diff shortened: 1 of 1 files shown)


DBMerger/Merger.cs
```diff
@@ -301,7 +307,7 @@ private void MergeUsers()
                     }
 
                     var oldAvatarTime = old.Length >= 3 ? (int)old[2] : 0;
```
loukylor, line 309: didn't realize avatar time could be null. could you also make sure this access of the row is not null? something like: `((int)old[2]) ?? 0` instead of `(int)old[2]` would work

loukylor (changes requested):

DBMerger/Merger.cs
book000: Probably because NullPointerException is raised at `((int)old[2])`, so `((int)old[2]) ? 0` would not solve the problem.
If the condition `old[2] ! = null to work around this.

loukylor: Ahh you're right, good catch. I'm low on sleep lol

loukylor (approved):

loukylor:
@Natsumi-sama Please merge when u can tyty

loukylor:
Ah also thanks for the pr. I hope my code wasn't too much of a nightmare to navigate

PR #1241, opened 2025-05-23
Groups Edit Mode Improvements

affectioned:
Added group bulk management UI

    You can now change visibility or leave multiple groups at once

    A "Select All" option allows quick bulk actions

    Each group now has an individual checkbox for selection

Fixed Vue warning: "<template v-for> key should be placed on the <template> tag"

    Rewrote the badge display logic to use a valid v-for on <el-popover>
DTtMiLAJnd


affectioned:
TODO: add some sleep delay between each API call, like 30 seconds when mass leaving more than 5 groups or something like that

PR #1284, opened 2025-07-07
Swap search and filter bar sizes to improve usability.

### LINKED ISSUE #1269: [Feature Request] Swap "Filter" and "Search" bars on some pages
On some pages, for example, the "Game Log" page, the Filter bar is a very wide bar in the center, while the search bar is a very small bar on the right.

This fact has caused me headache numerous times when trying to use the app with my VR overlays.

If these could be reversed, so the Search bar is the large bar in the middle, and the filter (which is just a dropdown button currently) was the smaller one on the side, it would help a lot. 

Also, since the filter bar may have N number of filters, it can probably expand leftward into the search bar if needed.

Image

nerdlan:
Building off of #1269 , I found it easier to navigate with a larger search bar than filters. I swapped the two and made the styling more consistent between views.

- **Search bars**: Enlarged to take primary space for easier interaction
- **Consistent layout**: Applied across all affected views

### Before Changes

| View | Screenshot |
|------|------------|
| **Feed** | image |
| **Game Log** | image |
| **Friend Log** | image |
| **Moderation** | image |
| **Notifications** | image |
| **Friend List** | image |

### After Changes

| View | Screenshot |
|------|------------|
| **Feed** | image |
| **Game Log** | image |
| **Friend Log** | image |
| **Moderation** | image |
| **Notifications** | image |
| **Friend List** | image |

### DIFF
--- a/src/views/Feed/Feed.vue
+++ b/src/views/Feed/Feed.vue
@@ -1,26 +1,18 @@
 <template>
     <div v-show="menuActiveIndex === 'feed'" class="x-container feed">
         <div style="margin: 0 0 10px; display: flex; align-items: center">
-            <div style="flex: none; margin-right: 10px; display: flex; align-items: center">
-                <el-tooltip
-                    placement="bottom"
-                    :content="t('view.feed.favorites_only_tooltip')"
-                    :disabled="hideTooltips">
-                    <el-switch v-model="feedTable.vip" active-color="#13ce66" @change="feedTableLookup"></el-switch>
-                </el-tooltip>
-            </div>
             <el-input
                 v-model="feedTable.search"
                 :placeholder="t('view.feed.search_placeholder')"
                 clearable
-                style="flex: 1 1 60%; margin-right: 10px"
+                style="flex: 1 1 30%; margin-right: 10px"
                 @keyup.native.13="feedTableLookup"
                 @change="feedTableLookup"></el-input>
             <el-select
                 v-model="feedTable.filter"
                 multiple
                 clearable
-                style="flex: 1 1 40%"
+                style="flex: 1 1 70%; margin-right: 10px"
                 :placeholder="t('view.feed.filter_placeholder')"
                 @change="feedTableLookup">
                 <el-option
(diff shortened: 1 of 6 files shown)


Map1en:
My idea is, "Search" is not a feature that most users often use. We should separate what is important from what is less important. After knowing what is important, the 6/4 ratio does not make sense. The order of the items should also be considered.

Also, the top bar design in tab page is a bit messy. For example, some tabs have a switch on the left and a refresh button on the right. If we want to change it, please consider these points and make sure each tab page has the same style.

nerdlan:
I see what you mean. A layout closer to this would be less messy.

### After Changes

| View | Screenshot |
|------|------------|
| **Feed** | image |
| **Game Log** | image |
| **Friend Log** | image |
| **Moderation** | image |
| **Notifications** | image |
| **Friend List** | image |

0xkubectl:
From a personal perspective, while I do not use the search often - when I use it, its one of the most valuable things VRCX has to offer and I would love to see it improved even beyond just the visual aspect.

Just some thoughts this PR:
- It might be a good idea to move the filter favourites button into the filters, as it effectively is one. This would remove visual clutter
- Fix the length of the filter + search bar to sth that is gonna stay consistent across different screens, i.e. have be so small that what ever is put to the left or right to it will not cause it to move. To put it visually, make it so search & filter align on every page.
grafik
- Maybe think about collapsing the filter menu into just a button that opens a dropdown, the icon of which would change color if a filter is active. This comes at the cost of not directly seeing what filters are active, not sure if this is worth it.
- possibly add a magnifying glass icon and filter icon (sth like this) to improve readability
grafik

Let me know what u think :]

Also with about ~12.5% of DAUs on VRCX just using the avatar search feature (based on the requests I recieve on avtrdb.com) I would assume search is popular enough the give it a thurough thought.

nerdlan:
Moving the filter favourites toggle inside the filter dropdown sounds like a good change. It could even be expanded to allow filtering between the individual groups of favorite friends, if possible.

PR #1287, opened 2025-07-12
Add group ban option to user profile

WeslieDE:
This adds a "Ban from group" option to the user profile menu (see images).
When you click this option, a dropdown menu will appear where you can select a group.
The dropdown is pre-filtered to only show groups where you have permission to ban users.
After selecting a group, you will be asked to confirm the action.

# Images

image
image
image

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -80,6 +80,7 @@ import FullscreenImageDialog from './components/dialogs/FullscreenImageDialog.vu
 import GalleryDialog from './components/dialogs/GalleryDialog.vue';
 import GroupDialog from './components/dialogs/GroupDialog/GroupDialog.vue';
 import InviteGroupDialog from './components/dialogs/InviteGroupDialog.vue';
+import BanGroupDialog from './components/dialogs/BanGroupDialog.vue';
 import LaunchDialog from './components/dialogs/LaunchDialog.vue';
 import PreviousInstancesInfoDialog from './components/dialogs/PreviousInstancesDialog/PreviousInstancesInfoDialog.vue';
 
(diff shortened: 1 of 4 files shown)


Map1en:
I personally think this feature doesn't match user habits.
When banning someone, users should select the user within the **group** page they want to manage, not choose a group from an individual user's dialog. Also, I'm unsure whether adding this feature is a good idea.

WeslieDE:
I know what you mean. But I don't think the option is so out of place. It's already possible to invite someone to the group, so why not kick them out again?

However, it's currently very impractical to ban someone from a group.
Currently, you have to search for the name, open the user profile, and copy the ID. Close each window, switch to the group, open the moderation tools, paste the ID there, and then you can ban them. That's a lot of clicking and takes forever, especially if you're dealing with several people at once.

But I would also be happy with a different way. The main thing is that it's faster and saves clicks.

Map1en:
image
I'm not completely sure if my answer is correct because I don't have access to any groups to try it out. But as far as I know, the Moderation Tools do have a bulk ban feature, right?

WeslieDE:
Yea it have but you cant enter names. You must past the id. So you are forced to open the user profile for every user to get this.

Map1en:
You’re right, I just tried it and it really doesn’t work. There are some issues with the group-related features because they’re hard to test.I think this is a small bug in the Moderation Tools. Fixing it will meet your needs and also help users who often use the Moderation Tools.
If you’re interested, submitting a PR would be the best way to help.
The user dialog code is already quite large and shouldn’t become the main page of VRCX.

Natsumi-sama:
I think enough people would benefit from this addition that it would be worth while adding, user dialog dropdown is already cluttered as is and this sure wont help but I see that as a separate issue.

Natsumi-sama:
Superseded with "

PR #1305, opened 2025-07-21
Improve 'Previous Instances' grouping behavior

osiris-plus:
Group a user’s previous instances by both location and calendar date for better clarity, instead of grouping only by instance identifier.

This issue is most noticeable with 24/7 public group instances, where the instance ID remains the same. In such cases, VRCX would group sessions from multiple days into a single row, making it much harder to track when you played with a particular user.

The algorithm connects multiple OnPlayerJoined / OnPlayerLeft events from gamelog_join_leave together if they're from the same instance and fall within 1 hour threshold window. This approach prevents log clutter from frequent rejoins as well.

Before:
image

After:
image

### DIFF
--- a/src/components/dialogs/UserDialog/PreviousInstancesUserDialog.vue
+++ b/src/components/dialogs/UserDialog/PreviousInstancesUserDialog.vue
@@ -165,7 +165,8 @@
         database.deleteGameLogInstance({
             id: props.previousInstancesUserDialog.userRef.id,
             displayName: props.previousInstancesUserDialog.userRef.displayName,
-            location: row.location
+            location: row.location,
+            localDate: row.localDate
         });
         removeFromArray(previousInstancesUserDialogTable.data, row);
     }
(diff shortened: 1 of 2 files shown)


Natsumi-sama:
We cannot split instances by date in this way because people can be in an instance across 2 days, e.g. from 11pm to 2am. A proper solution would be checking for next location date and limiting the scope of a single instance till next unique location is hit. Might even need a threshold in there to support rejoining from home world after a crash.

osiris-plus:
I've committed the changes.

The algorithm now connects multiple OnPlayerJoined / OnPlayerLeft events from gamelog_join_leave together if they're from the same instance and fall within 1 hour threshold window. This takes care of rejoins as well.

Also the dates are now correct and reflect the OnPlayerJoined date.

osiris-plus:
Forgot screenshot:

image

Although there are more records compared to the previous iteration, the data is more accurate and therefore more useful.

PR #1315, opened 2025-08-04
slightly modernized & updated default dark theme

affectioned:
Theme color could be adjusted back to blue or maybe darker blue? I just picked this purplish, because I thought it would give nice constrast, but feel free to change it however you like. It is inspired by YouTube dark theme color pallete. Thanks for the refactor, made it quite easy to work with.
Screenshot 2025-08-04 213739
Screenshot 2025-08-04 213723
Screenshot 2025-08-04 213956

### DIFF
--- a/src/assets/scss/themes/theme.amoled.scss
+++ b/src/assets/scss/themes/theme.amoled.scss
@@ -11,22 +11,24 @@
 $--theme-hue: 0;
 $--theme-saturation: 0%;
 
-$--theme-bg-1: hsl($--theme-hue, $--theme-saturation, 6%); // #101010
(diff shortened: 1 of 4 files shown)


Map1en:
I think you should add a new theme instead of modifying the base dark theme.
[Wiki - Themes](https://github.com/vrcx-team/VRCX/wiki/Themes)

affectioned:
Sure, could I call it Dark (New) or something along those lines?

Map1en:
I personally think Dark （New） is not appropriate. It would be better to follow the existing dark themes. Or come up with a unique name.

affectioned:
updated it, named it amoled since it's quite dark

PR #1352, opened 2025-08-26
Fix launchcommands on Linux

0xkubectl:
Removed reliance on the old IPC server, use current electron best practices to make launch comamnds work on all platforms. Please verify if I have used the primitives correctly and did not create a security oopsie. 

[Here](https://github.com/user-attachments/assets/f74b38f3-45f3-48bc-8b51-b131ec55f6f9) is a demo of how it looks.

This requires testing on Windows & mac OS.

### DIFF
--- a/src-electron/main.js
+++ b/src-electron/main.js
@@ -14,7 +14,18 @@ const { spawn, spawnSync } = require('child_process');
 const fs = require('fs');
 const https = require('https');
 
+const VRCX_URI_PREFIX = "vrcx"
+
 //app.disableHardwareAcceleration();
+
+if (process.defaultApp) {
+    if (process.argv.length >= 2) {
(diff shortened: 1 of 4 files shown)


0xkubectl:
Sadly this implementation has a bug.
When the eventloop is properly initialized, the events are properly handled, but while launching they are not.
What I tried:
- checking argv and if sth matches the prefix sending it to the handler, apparently also does not work - i have no clue then this needs to be sent
- i thought it was related to the login state,  so tried waiting for it to go true, but the events are not even routed.

Maybe someone has an idea on how to approach this.

Natsumi-sama:
In dotnet side of the code it already handles launch args during startup so don't need to worry about that.
I've adding back Cef support in my Lint commit.
Tested and working well under Electron on Windows so I'll assume it'll work fine on Linux, as for macOS, it'll either work or it wont I'm not willing to test myself.

PR #1369, opened 2025-09-11
Open gallery popup when redirecting to tools tab

Cayahuanca:
The Gallery has been moved to the Tools page, but if simply open the Tools page when clicking "Manage VRC+ Images & Inventory" from my profile, it will have to click "Manage VRC+ Images & Inventory" in tools page again to open it, which is a double step.
To ensure that users are aware that the page has moved, I think it would be better to redirect them to the Tools page but keep the pop-up displayed as usual, ensuring the convenience of using the previous operations.

Animation


Map1en:
The next stable version will probably get rid of these old buttons.❤️

Natsumi-sama:
This is one of those rare scenarios where it being less convenient is more ideal, it encourages the user to use the new buttons over the old buttons that take more clicks, once the old buttons are removed they will already be comfortable navigating using the new buttons.

I'll merge the other PRs once the merge conflicts are resolved.

PR #1414, opened 2025-10-12
fix: prevent default autoUpdateVRCX config from being unexpectedly suppressed

xuan25:
fix: prevent default autoUpdateVRCX config from being unexpectedly suppressed

For OS dstros that should not rely on a 3rd-party updater,  a reasonable patch like 
```
--- a/src/stores/vrcxUpdater.js
+++ b/src/stores/vrcxUpdater.js
@@ -18,1 +18,1 @@
-        autoUpdateVRCX: 'Auto Download',
+        autoUpdateVRCX: 'Off',
```
should do the job. For example, for Arch Linux, people can apply a patch like this https://aur.archlinux.org/cgit/aur.git/tree/build.patch?h=vrcx

However, the default value in `state.autoUpdateVRCX` was suppressed by `configRepository.getString('VRCX_autoUpdateVRCX', '')`. This causes the default value in `state.autoUpdateVRCX` to be ignored and potentially misleading, which can confuse people and make the issue harder to trace.

### DIFF
--- a/src/stores/vrcxUpdater.js
+++ b/src/stores/vrcxUpdater.js
@@ -48,7 +48,7 @@ export const useVRCXUpdaterStore = defineStore('VRCXUpdater', () => {
         }
 
         const [autoUpdateVRCX, vrcxId] = await Promise.all([
-            configRepository.getString('VRCX_autoUpdateVRCX', 'Auto Download'),
+            configRepository.getString('VRCX_autoUpdateVRCX', state.autoUpdateVRCX),
             configRepository.getString('VRCX_id', '')
         ]);
 


Natsumi-sama:
That if statement is only meant for handling old config values that don't exist anymore, as Linux users aren't the primary users having auto updater enabled by default is very much intended, however we could add code to default it off for Linux users, maybe that would be more appropriate, although in saying that I'm sure most Linux users are using the AppImage rather than their package managers.
Side note, package maintainers should not be needing to patch code, that's just silly, we could add a launch arg or VRCX.json config line, would be much more appropriate.

xuan25:
Thank you for sharing your thoughts.

1.

I **agree** with the design of **enabling updates** by default.

However, the current issue is that the code that **handles legacy config** will always override the **default behavior** declared in the `state`, which means the **default behavior** declared in the `state` never actually takes effect. In other words, the software’s **default behavior** is entirely determined by the code that processes old configurations; the default values declared in the `state` do not control the defaults.

Regarding the two hard-coded instances of `“Auto Download”`, I think keeping one of them as the default behavior is reasonable. Compared with using the **legacy-config handling** to control the default, I believe it’s more appropriate to control the default in the **`state` initialization**.

2.

As for the patch, this is a documented, standard practice in my OS distro’s packaging pipeline (the Arch build system) and is quite common. 

It allows the main repository to focus on core feature development without needing to account for the differences among all package managers, OS distros and their **release cycles** (system dependencies may be updated frequently). Therefore, I don’t have strong opinions about whether doing it this way is appropriate.

Natsumi-sama:
There is no bug here, I think you're misunderstanding the code, `configRepository.getString` will return config value from DB, if no DB value is set, it will fallback to default value, `autoUpdateVRCX` is now set with with either DB value or default value. `if (autoUpdateVRCX === 'Auto Install')` will almost never be true, now state matches value from DB with `state.autoUpdateVRCX = autoUpdateVRCX;`, the set state default (`autoUpdateVRCX: 'Auto Download',`) is only used as an intermediary value before config has been loaded, it could be set to anything really it wouldn't matter, but having a sane value in there makes the most sense.

Avoiding upstreaming proper fixes/changes that aren't distro specific makes zero sense to me.

xuan25:
**A) If** you take the DB querying returned value as a placeholder/intermediate value, then:

If no DB value is found and a placeholder is returned, the `state` should keep its initial (default) value isn't it? The current implementation will always set the `state` value to the placeholder if no DB value is found, which means the initial (default) value in `state` will never be used. In this way, I believe the logic was wrong.

I believe an alternative is to replace
```c#
configRepository.getString('VRCX_autoUpdateVRCX', 'Auto Download')
...
if (autoUpdateVRCX === 'Auto Install'){
    ...
}
```
with
```c#
configRepository.getString('VRCX_autoUpdateVRCX', state.autoUpdateVRCX)
...
if (state.autoUpdateVRCX === 'Auto Install'){
    ...
}
```
with initialization in `state` as is
```c#
const state = reactive({
    ...
    autoUpdateVRCX: 'Auto Download',
    ...
});
```

**B) Or, if** you believe the default behaviour should be stated close to the DB querying code, then I think we should keep the DB querying as is, and remove the misleading initialization in `state`:
```c#
const state = reactive({
    ...
    autoUpdateVRCX: '', // or null
    ...
});
```

---

The default behaviour should either be controlled in `state` initialization, or on the DB querying code, but not both, isn't it? In either case, the current implementation is wrong or misleading.

Natsumi-sama:
Both methods and existing method will all end in same result, sure I'd say method A is a cleaner approach, but in the grand scheme of things it's not a big deal.

Natsumi-sama:
Issue with changing this one line is now it not longer matches the standard across the entire program, every `configRepository` related line in the program would need to be changed for this to be worthwhile, and there is quite a lot, it isn't worth it.

xuan25:
I have no clue which standard you’re referring to. Could you please elaborate, especially where the default behaviour of the program should be defined/declared according to the standard?

Natsumi-sama:
This bit of code is repeated throughout the entire program, changing it in only one place would make it inconsistent, so this change needs to happen for the entire code base or not at all.

PR #1416, opened 2025-10-13
Fixed Material 3 theme

poprox24:
Fixed the material 3 theme, tested at 1080p and 1440 with different zoom amounts

Preview:
image
image
image
image
image
image
image
image

### DIFF
--- a/src/assets/scss/themes/theme.material3.scss
+++ b/src/assets/scss/themes/theme.material3.scss
@@ -939,7 +939,8 @@ input[type='number'],
     position: relative;
 }
 
-.el-button--small.is-circle {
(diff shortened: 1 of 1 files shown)


poprox24:
A note: I plan to maintain this theme from now on unless someone else is up for the job, for me this is the only theme I can use VRCX with

Natsumi-sama:
Awesome, thanks for the help, use nightly builds if you aren't already to keep on top of changes as they happen.

poprox24:
> Awesome, thanks for the help, use nightly builds if you aren't already to keep on top of changes as they happen.

Ah yeah thanks, would have forgotten to change that

PR #1463, opened 2025-11-03
[Material 3] Friend Location Update

poprox24:
https://github.com/user-attachments/assets/54005fe8-6ec7-4231-80b6-e82085a1f5f1

I will be able to show the same instance list, once I have some friends there, I checked before and it looks good but I cannot take a picture right now
But without it for now, the instance name is set to be white-ish (--md-sys-color-on-background)


poprox24:
image

Changed the top buttons to fit the theme properly, fits much more
Don't see anything else that doesn't fit right now so I won't be making any more changes

poprox24:
image
Finally able to make a full preview here ^^

poprox24:
Also noticed I could make the line dividing the online friends match the material 3 scheme so I will do that right now

PR #1479, opened 2025-11-10
Adding keyboard shortcut for direct access + Fixes for Linux clipboard reading

0xkubectl:
- **feat: add ctrl+d keyboard shortcut to open direct access**
- **fix: make sure direct access promt only get shown once**
- **fix: linux clipboard acess on non x11 systems**

There was some debate on where to put the direct access button, so I wanted to make it independent of where the button is to have quick and easy access to it.
This PR adds the `Ctrl+D` shortcut when VRCX is focused, which will call the same function which is called when the Direct Access button is clicked.

I also added a fix that when the dialog is open, that you will not be able to open it multiple times and fixed the Linux clipboard not being read on non-x11 machines (i.e. machines that do not have xclip)

Let me know what you think.
Also feel free to just cherry-pick the linux clipboard fix.

### DIFF
--- a/src-electron/main.js
+++ b/src-electron/main.js
@@ -11,6 +11,7 @@ const {
     nativeImage
 } = require('electron');
 const { spawn, spawnSync } = require('child_process');
+const { clipboard } = require('electron');
 const fs = require('fs');
 const https = require('https');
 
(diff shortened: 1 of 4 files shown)


Map1en:
That's a good idea, but I think most users aren't even aware of the existence of direct access, let alone the keyboard shortcuts.

0xkubectl:
I personally could think of three people who would actually enjoy that feature, but I'm not personally attached to this, so we can rename this PR to just include the Linux clipboard fixes. Though I would like to run it by some people on the discord first.

PR #1480, opened 2025-11-11
ui: popover anchored to hovered UI element

0xkubectl:
linked to this discussion in DC: https://discord.com/channels/854071236363550763/864248845301710908/1437520101866995795

- added function that will instantly hide already open popover if a different one is hovered
- increased desktop hide-after timeout to 250
- added fixed background for VR
- decreased the height of the popover for popover content to be closer to the hovered element


0xkubectl:
Currently on hold as I am busy with other things

PR #1482, opened 2025-11-11
Keyboard Shortcut Improvements

0xkubectl:
- **fix: add useMagicKeys from vueuse**
- **fix: use useMagicKeys composable to watch for keyboard changes**

~~Use a more composable way of accessing keyboard shortcuts. Stolen & adapted from [useMagicKeys](https://github.com/vueuse/vueuse/blob/main/packages/core/useMagicKeys/index.ts) from [vueuse](https://github.com/vueuse/vueuse).~~

Also I tried adding the `KeyD` keycode to hopefully fix some windows versions that ate glue for breakfast. I have not tested that, it should not regress for already working versions. 

Edit:
I added @vueuse/core as a dependency - there is a bunch of useful stuff in there, that will probably be helpful elsewhere. Furthermore I took the two instances where keyboard interactions where listened to and moved them to useMagicKeys. While testing the shortcuts that open/reload the window I noticed the devtools where not opening on Linux and added a fix for that aswell.

Please let me know what you think :]

### DIFF
--- a/src-electron/main.js
+++ b/src-electron/main.js
@@ -275,6 +275,10 @@ ipcMain.handle('app:setTrayIconNotification', (event, notify) => {
     setTrayIconNotification(notify);
 });
 
+ipcMain.handle('app:openDevTools', () => {
+    if (mainWindow) mainWindow.webContents.openDevTools();
+});
+
 function tryRelaunchWithArgs(args) {
     if (
         process.platform !== 'linux' ||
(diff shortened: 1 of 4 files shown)


0xkubectl:
integrated in https://github.com/vrcx-team/VRCX/commit/39a8685f0d299ee81234d7ff6ea6724e2973a664

Issue #1496, opened 2025-11-18
[Feature Request] Show the whole ID for deleted/private favorites


RobSkop:
When browsing the favorites you can not see the whole world/user IDs when the world or user got deleted or whatever, after the latest update.

It would be good to be able to show the whole ID since clicking it will open the world dialog briefly before throwing the error that the world is deleted/private.

It was always good to be able to copy the world-ID to check online which world it was, now that is not really possible anymore.

PR #1523, opened 2025-12-06
Feature: Add 'Copy ID to clipboard' to world load error.

GapVR:
Append "Copy ID to clipboard" link to "Failed to load world." error message.

Resolves https://github.com/vrcx-team/VRCX/issues/1496.

### DIFF
--- a/src/stores/world.js
+++ b/src/stores/world.js
@@ -132,11 +132,26 @@ export const useWorldStore = defineStore('World', () => {
             .catch((err) => {
                 D.loading = false;
                 D.visible = false;
+                const errWorldId = L.worldId;
+                const errPopupId = `worldRequest-error-${Date.now()}`;
+                const errPopupMsg = `Failed to load world. <span id="${errPopupId}" style="color: #50D0FF; cursor: pointer; text-decoration: underline;">Copy ID to clipboard</span>`;
                 ElMessage({
-                    message: `Failed to load world. <a href="#" onclick="navigator.clipboard.writeText('${L.worldId}'); event.preventDefault();">Copy ID to clipboard</a>`,
+                    message: errPopupMsg,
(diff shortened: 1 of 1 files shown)


Natsumi-sama:
This is not a safe enough way of handling this, it's asking for an XSS due to the many different ways users can open worldDialog, from URI/browser extension to bio links, direct access, etc.

GapVR:
It should be safer now.

dangerouslyUseHTMLString is used only to reference ${Date.now()} to make a unique element to handle clicks.

Natsumi-sama:
I more meant by using a more appropriate component like `ElMessageBox`, and if I were to add this I'd add it for avatars, users, groups and worlds.

GapVR:
I think the current ElMessage toast is best since it's the least disruptive. If you want, I could try to add this as a function to ../shared/utils and make avatars, users, groups and worlds use it at "Failed to load...".

Natsumi-sama:
hmm yea I kind of agree it is a bit more disruptive than I'd like but I don't want to add HTML in this way.

Natsumi-sama:
Here an example of how to better handle this inline HTML https://github.com/vrcx-team/VRCX/pull/1525/files
I'll also make it so favorite IDs aren't getting chopped off so they can be copied from there as well.

Map1en:
Inserting HTML is definitely out of the question. Since we have better methods available, we shouldn't rely on such a risky approach.

Furthermore, using `ElMessage.error` for this interaction results in poor UX. It defaults to a 3-second duration, and extending that time specifically for the copy action would likely confuse users who don't need that functionality.

Therefore, I believe this PR requires further discussion and revision.

GapVR:
Thanks for the feedback. I will close this PR and leave it to the professionals.

PR #1549, opened 2025-12-27
various bundle optimizations

ariesclark:
- missing "@element-plus/icons-vue" dependency
- update vite (40% faster builds)
- don't include sentry in non-nightly builds
- swap to variable fonts & don't include font files in repo (82% smaller)
- lazy load languages to not keep them in memory

### DIFF
--- a/src/plugin/i18n.js
+++ b/src/plugin/i18n.js
@@ -4,7 +4,6 @@ import { getLocalizedStrings, languageCodes } from '../localization';
 const i18n = createI18n({
     locale: 'en',
     fallbackLocale: 'en',
-    availableLocales: languageCodes,
     legacy: false,
     globalInjection: false,
     missingWarn: false,
--- a/src/stores/settings/advanced.js
+++ b/src/stores/settings/advanced.js
@@ -12,13 +12,14 @@ import { watchState } from '../../service/watchState';
 
 import configRepository from '../../service/config';
 import webApiService from '../../service/webapi';
+import { languageCodes } from '../../localization';
 
 export const useAdvancedSettingsStore = defineStore('AdvancedSettings', () => {
     const gameStore = useGameStore();
     const vrcxStore = useVrcxStore();
     const VRCXUpdaterStore = useVRCXUpdaterStore();
 
-    const { availableLocales, t } = useI18n();
+    const { t } = useI18n();
 
     const state = reactive({
         folderSelectorDialogVisible: false
@@ -163,7 +164,7 @@ export const useAdvancedSettingsStore = defineStore('AdvancedSettings', () => {
 
         if (
             !bioLanguageConfig ||
-            !availableLocales.includes(bioLanguageConfig)
+            !languageCodes.includes(bioLanguageConfig)
         ) {
             bioLanguage.value = 'en';
         } else {
--- a/src/stores/settings/appearance.js
+++ b/src/stores/settings/appearance.js
@@ -24,6 +24,7 @@ import { useVrcxStore } from '../vrcx';
 import { watchState } from '../../service/watchState';
 
 import configRepository from '../../service/config';
+import { languageCodes } from '../../localization';
 
 export const useAppearanceSettingsStore = defineStore(
     'AppearanceSettings',
@@ -37,7 +38,7 @@ export const useAppearanceSettingsStore = defineStore(
         const router = useRouter();
         const uiStore = useUiStore();
 
-        const { t, availableLocales, locale } = useI18n();
+        const { t, locale } = useI18n();
 
         const MAX_TABLE_PAGE_SIZE = 1000;
         const DEFAULT_TABLE_PAGE_SIZES = [10, 15, 20, 25, 50, 100];
@@ -182,7 +183,7 @@ export const useAppearanceSettingsStore = defineStore(
 
                 const lang = result.split('-')[0];
 
-                for (const ref of availableLocales) {
+                for (const ref of languageCodes) {
                     const refLang = ref.split('_')[0];
                     if (refLang === lang) {
                         await changeAppLanguage(ref);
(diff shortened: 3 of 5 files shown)


Map1en:
Thanks for your contribution! I've reviewed the changes and there are a few points that need to be addressed before we can move forward:

1. #### Vite Version:

   - We should not use **Vite beta versions** for our stable releases. The Vue ecosystem and its dependencies iterate very rapidly, using a beta version introduces unnecessary maintenance risks and stability concerns that we'd like to avoid. Please downgrade to the latest stable version.

2. #### i18n & Localization Issues:

   - Regression: The language selection in the settings is currently broken. It seems the language list is no longer being populated or read correctly.

   - File Format: Please revert the i18n files from `.js` back to `.json`. We specifically use JSON to make it easier for external translators to contribute. Moving to JS files makes it much harder for non-technical contributors to manage translations.

3. #### Bundle Analysis:

   - We do not need a permanent bundle analysis library in the repository. For a desktop application, the overhead of maintaining this library isn't justified. Please remove it from the dependencies.

Please look into these issues. Let me know if you have any questions!

ariesclark:
> We should not use Vite beta versions for our stable releases. The Vue ecosystem and its dependencies iterate very rapidly, using a beta version introduces unnecessary maintenance risks and stability concerns that we'd like to avoid. Please downgrade to the latest stable version.

Vite's v8 beta is considered pretty stable, with [quite remarkable performance gains](https://github.com/vitejs/rolldown-vite-perf-wins). Internally, [the biggest change was swapping from Rollup to Rolldown](https://main.vite.dev/guide/migration#rolldown) (their Rust transpiler) that provides almost 1-1 guarantees. I can understand if you would prefer to wait for their stable release though, will downgrade.

> Regression: The language selection in the settings is currently broken. It seems the language list is no longer being populated or read correctly.

Must've broken when I merged upstream changes, will take a look.

> File Format: Please revert the i18n files from .js back to .json. We specifically use JSON to make it easier for external translators to contribute. Moving to JS files makes it much harder for non-technical contributors to manage translations.

You mean translators directly editing the bundled files? Is that the ideal workflow you want to support? I wonder if there's a cleaner workflow for translators. Will look into this.

> We do not need a permanent bundle analysis library in the repository. For a desktop application, the overhead of maintaining this library isn't justified. Please remove it from the dependencies.

Sure, but I don't see how it introduces any "overhead of maintaining this library".

ariesclark:
I have resolved the issues you mentioned.

Map1en:
Thanks for the explanation. I’ve organized my thoughts below.

1. #### Vite
I know about these changes and the performance gains, but reducing build time is practically meaningless for us since we don’t have a continuous CI/CD workflow. On the other hand, introducing a beta version where the entire underlying logic has swapped to Rolldown/OXC brings huge risks. I don't think any project or company would use a beta library in a stable release especially for something as critical as the bundler.

2. #### i18n
We used to have Crowdin, but we don't anymore. Now all translations are submitted through PRs, so keeping them as JSON is the simplest and most direct way for everyone. Also, in your current modified version, the i18n text doesn't update even after modifying the files and refreshing the app. We need to revert this to ensure the translation workflow stays functional.

3. #### Bundle Analysis
You're right, there isn't much "maintenance overhead," I probably put that too strongly. However, for a desktop app, we don't need to optimize every single KB; keeping track of changes at the MB level is more than enough. The chances of us actually needing this library are very low, so it’s better to just remove it.

ariesclark:
> Also, in your current modified version, the i18n text doesn't update even after modifying the files and refreshing the app. We need to revert this to ensure the translation workflow stays functional.

Hot-reload does work for translations. The exception is the language name itself, which is intentionally bundled at build time to avoid loading every language file just to display the language selector. Let me know if it's still not updating for you and we can look into it.

8e96a081392a39c7

> You're right, there isn't much "maintenance overhead," I probably put that too strongly. However, for a desktop app, we don't need to optimize every single KB; keeping track of changes at the MB level is more than enough. The chances of us actually needing this library are very low, so it’s better to just remove it.

For what it's worth, this PR cuts 80MB, not KB. I get that a desktop app doesn't need to optimize every byte, but bundle size shouldn't be quietly ignored, especially when the savings are meaningful. I've already removed the analyzer.

Let me know if you run into anything else.

BenjaminZehowlt:
I get the reason for changing font for optimization, I looked at Fontsource. It's the font I thought that's always been a bit off and could never put a finger on it. Just personal input, also changing font needs determination since its the base theme.

Map1en:
> Hot-reload does work for translations. The exception is the language name itself, which is intentionally bundled at build time to avoid loading every language file just to display the language selector. Let me know if it's still not updating for you and we can look into it.

Yes, translations update without issues in the dev environment. However, some contributors have told us that many of them don't know how to use Node to build the frontend, let alone use VS or Rider for the C# side. It's straightforward for us as developers, but it’s a high barrier for them. That’s why we have this convenient method of directly editing the JSON in the build artifacts for translations.

Regarding the font issue, I identified and fixed it a while ago, but it's on a different branch and hasn't been merged yet. My current approach was just putting the font files in the repo, which isn't ideal. I really like your idea of filtering them during the build process completely agree with that.

ariesclark:
> Yes, translations update without issues in the dev environment. However, some contributors have told us that many of them don't know how to use Node to build the frontend, let alone use VS or Rider for the C# side. It's straightforward for us as developers, but it’s a high barrier for them. That’s why we have this convenient method of directly editing the JSON in the build artifacts for translations.

I'm pretty sure you misinterpreted what I said, the video I shared is a production build, not the development environment.

Map1en:
> I'm pretty sure you misinterpreted what I said, the video I shared is a production build, not the development environment.

Oh, my mistake. It actually works. Then that works for me.

PR #1583, opened 2026-01-17
feat: added striped table mode for visual clarity

0xkubectl:
Added striped tables.
Added a toggle under apperance to revert to original behavior.

Translation missing for anything but english for that settings item.

## Caveat
This value is read from the apperanceStore inside of the DataTable, not sure if there is a significant performance penalty involved in doing it this way. If so, I will adapt the patch to use the method used for compact tables.

Let me know what you think!


Map1en (approved):
Reviewed and approved.
Looks good to me.

0xkubectl:
Superceded by #1585

PR #1587, opened 2026-01-19
Update Japanese Translation

Cayahuanca:
Update Japanese Translation
Fix Notification.vue that translation was not reflected in the "Confirm" section.

image


Map1en:
お疲れ様です～❤️

PR #1595, opened 2026-01-24
Colored private avatars lock icon

monolithic827:
The white lock icon blends in when scanning through multiple entries, making it difficult to quickly understand that an avatar is private.
Adding color to the icon improves glanceability, similar to how it was handled in the previous UI.


Map1en:
At the moment, only private avatars use a lock icon, which is already clear enough.
Additionally, introducing unnecessary colors into a table with a large amount of information would distract visual focus, so we’ve decided not to merge this PR.

PR #1596, opened 2026-01-24
fix: AES key generation out of bounds

wilk-polarny:
AES key generation assumes that character amount = byte length, which is a dangerous assumption when working with multi-byte characters.
Using a password with multi-byte characters might trigger the padding logic, if the password has less than 32 characters but more than 32 bytes, thus leading to illegal Array accesses.
The proposed solution is to go off of the actual byte array, rather than the character sequence.


BenjaminZehowlt:
Honestly, a better fix here would be to change it to hash the password with sha256 since it'll always be exactly 32 bytes (256 bits)

Yes, I'm aware this would break existing master passwords.

wilk-polarny:
Yea, the proposal is just a quick fix.
The whole key generation and storage needs a refactor, since the current approach including the padding with the static string is a bit flaky and opens doors to a lot of other issues (including weakening of the encryption for short passwords).

Obviously, that also requires a migration strategy to migrate existing encrypted passwords for users that still have primary passwords enabled (new default is OFF). Whether a stored account password is encrypted or not is currently only determined by a global boolean flag.
I'd propose to bake this info into each individual stored password (or as part of the data structure for saved accounts) defining whether it's encrypted, and which method (no info = legacy = trigger migration) it uses.
I'd suggest to apply encryption to the session tokens (cookies) too when using a primary password.

BenjaminZehowlt:
As a temporary workaround this is fine, but long term that is 100% something we need to do. Probably be best to tie it to install specific information.

PR #1597, opened 2026-01-25
feat(search): support prefix-based user and world ID search

WavesMan:
- Add searchUserByIdPrefix to enable quick lookup by user ID prefix (usr_).
- In the search view, when the query starts with usr_ or wrld_, first match cached records whose IDs start with the prefix.
- If no cached matches and the ID length is a full UUID, attempt to fetch the entry via API.
- Preserve existing name/tag search behavior for non-prefixed queries.
- Improves efficiency when searching with known ID fragments.

### DIFF
--- a/src/stores/search.js
+++ b/src/stores/search.js
@@ -4,7 +4,7 @@ import { toast } from 'vue-sonner';
 import { useI18n } from 'vue-i18n';
 import { useRouter } from 'vue-router';
 
-import { compareByName, localeIncludes } from '../shared/utils';
+import { compareByName, compareByDisplayName, localeIncludes } from '../shared/utils';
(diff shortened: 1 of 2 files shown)


Natsumi-sama:
By chance, have you missed our "direct access" button?

Map1en:
左侧导航栏有个直接打开按钮，粘贴进去或者先复制，就能直接打开这些。
'' - Translated from an LLM of unknown.

WavesMan:
I used the old version before and didn't notice this new option, but I think it can also be added as an additional supplement

Map1en:
我觉得没太大必要在search里再加一遍，sry。

PR #1612, opened 2026-01-29
optimize vr runtime detection

sofoxe1:
idk c# and didn't test since i don't use this feature but similar implementation works here: https://codeberg.org/sofoxe1/xr_overlay/src/commit/614c687827913e9585a432d62b9e0eae7b976e99/core/src/xr/mod.rs#L234
i will have to look at how steamvr works to avoid process scanning when checking for it

### DIFF
--- a/Dotnet/AppApi/Electron/GameHandler.cs
+++ b/Dotnet/AppApi/Electron/GameHandler.cs
@@ -31,10 +31,8 @@ namespace VRCX
 
         public override bool IsSteamVRRunning()
         {
-            var processNames = new[] { "vrmonitor", "monado-service" };
-            foreach (var name in processNames)
             {
-                var processes = Process.GetProcessesByName(name);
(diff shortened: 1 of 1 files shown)


Natsumi-sama:
`IsSteamVRRunning` is only used on Linux, is `Process.GetProcessesByName()` really *that bad* on Linux? I feel like this PR would introduce bugs more than anything else. I'd like to get some input from other Linux users if they can test how `Process.GetProcessesByName()` performs.

sofoxe1:
the only issue it causes is the cpu usage number in the process monitor, as for bugs depends what exactly vrcx should detect as running for example for application to be able to start under wivrn both config and pipe needs to be present but it needs the config only when loading openxr and wivrn could also randomly crash without cleaning those files

Map1en:
Sorry, we’d rather avoid introducing potential bugs. The current version is relatively stable, so we’ll consider it once code quality can be reliably ensured. For now, we’ll close this.

PR #1619, opened 2026-02-01
Make avatar thumbnail display consistent with other images

LunarKittyy:
Avatar thumbnails in the dialog were displaying without `object-fit: cover`, causing them to appear squished when the source image didn't match the 4:3 aspect ratio of the container.

This adds `object-fit: cover` to match the behavior of all other image thumbnails in the app.

### DIFF
--- a/src/components/dialogs/AvatarDialog/AvatarDialog.vue
+++ b/src/components/dialogs/AvatarDialog/AvatarDialog.vue
@@ -13,7 +13,7 @@
                         :src="avatarDialog.ref.thumbnailImageUrl"
                         class="cursor-pointer"
                         @click="showFullscreenImageDialog(avatarDialog.ref.imageUrl)"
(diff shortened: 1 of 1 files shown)


Natsumi-sama:
I had originally done this intentionally, for when people upload photos to VRCX, it accurately shows how they will look in-game and on the website.

LunarKittyy:
I went through the website and in-game to double-check, and it looks like VRChat has actually updated most of their UI to avoid squishing. The avatar detail pages now use `object-fit: cover`, the friends list and world cards use `background-size: cover`, and even the search results crop properly.
The only place I found that still squishes is the "Current Avatar" preview at the top of the Avatars page, and maybe the Favorites page, both of which still use `object-fit: fill` in fixed 4:3 containers. But those seem like legacy holdovers at this point and affect only the website.

Natsumi-sama:
Thanks for giving it a thorough look over, saves us the effort.

PR #1620, opened 2026-02-01
fix typo

eeacks:
This is a small error I found while exploring the source code.
I'd like to take some time to discuss this. I'm trying to implement a VRCX-compatible API as a provider, but I can't seem to find any documentation on how to properly implement a VRCX-compatible API. I'm not sure if the lack of documentation is because this is a gray area.
Although this source file explains everything, some other providers seem to have returned additional information during implementation.

### DIFF
--- a/src/stores/avatar.js
+++ b/src/stores/avatar.js
@@ -454,7 +454,7 @@ export const useAvatarStore = defineStore('Avatar', () => {
             console.error('Failed to get avatar images:', error);
             return {
                 ownerId: '',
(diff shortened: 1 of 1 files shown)


Natsumi-sama:
If you DM me on Discord I can help with any info you may need, providers only need to return the normal VRC avatar JSON, some return some extra data and some return the wrong object key names (skill issue). We should really document it though.

eeacks:
got it, i've sent a request

PR #1625, opened 2026-02-02
OpenAI api add  Fetch Models

gujimy:
1.Added a Fetch Models button for OpenAI-compatible providers to load available models from custom endpoints and select them from a dropdown.

2.Added world description translation support .

### DIFF
--- a/src/components/dialogs/WorldDialog/WorldDialog.vue
+++ b/src/components/dialogs/WorldDialog/WorldDialog.vue
@@ -150,14 +150,18 @@
                                 </Badge>
                             </template>
                         </div>
-                        <div style="margin-top: 5px; display: flex; align-items: center;">
+                        <div style="margin-top: 5px; display: flex; align-items: center">
                             <span
                                 v-show="worldDialog.ref.name !== worldDialog.ref.description"
-                                style="font-size: 12px; flex: 1; margin-right: 0.5em;"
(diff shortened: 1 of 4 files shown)


Map1en:
我看了一下，有几个问题可能需要修改一下。

1. 因为现在用了tailwind，修改样式要用tailwind的class来改。
2. Translated by...这个不需要，user dialog那边也有问题到时候我们会改下。
3. i18n的`dialog.world.info.translate_description`没用上。
4. 拼接`/models`的时候根据用户输入的值可能出现两遍`/models`，或者query没带上。
5. `fetchAvailableModels`里面已经有toast.error了，外部调用的`fetchModels`里还有一份，会弹两个。
6. 提交前使用vscode的的esbenp.prettier-vscode插件format一下。

I looked a bit, have several problems maybe need change a bit.

1. Because now used tailwind, change style must use tailwind's class to change.
2. Translated by... this no need, user dialog there also have problem, that time we also change.
3. i18n's `dialog.world.info.translate_description` not used.
4. When splicing `/models`, according to user input value maybe appear two times `/models`, or query not 5. bring on.
5. Inside `fetchAvailableModels` already have toast.error, outside calling `fetchModels` inside still have one, will pop two.
6. Before submit use vscode's esbenp.prettier-vscode plugin format a bit.

gujimy:
**The code has been updated. Please check.**

PR #1632, opened 2026-02-06
add cute avatar search provider

imZuxi:
Since Just H is most likely not coming  back around i thought id add the same search the world [Cute Avatar Search](https://vrchat.com/home/world/wrld_15c6ff42-a779-40a2-8c30-862d9015795e) uses, edit this can also be used as a backup in-case avtrdb fails

### DIFF
--- a/src/stores/avatarProvider.js
+++ b/src/stores/avatarProvider.js
@@ -15,7 +15,8 @@ export const useAvatarProviderStore = defineStore('AvatarProvider', () => {
     const avatarRemoteDatabaseProvider = ref('');
 
     const avatarRemoteDatabaseProviderList = ref([
-        'https://api.avtrdb.com/v3/avatar/search/vrcx'
+        'https://api.avtrdb.com/v3/avatar/search/vrcx', 
+        'https://avtr.icu/search'
     ]);
     watch(
         () => watchState.isLoggedIn,
@@ -29,7 +30,7 @@ export const useAvatarProviderStore = defineStore('AvatarProvider', () => {
         avatarRemoteDatabaseProviderList.value = JSON.parse(
             await configRepository.getString(
                 'VRCX_avatarRemoteDatabaseProviderList',
-                '[ "https://api.avtrdb.com/v3/avatar/search/vrcx" ]'
+                '[ "https://api.avtrdb.com/v3/avatar/search/vrcx", "https://avtr.icu/search" ]'
             )
         );
         const deprecated = 'https://avtr.just-h.party/vrcx_search.php';


Misaka-L:
Unable to connect to `https://avtr.icu/search` due to anti-bot. By the way, do you the owner of the world or search provider?

Provini:
According to your website, shouldn't this be https://avtr.zuxi.dev/search? At least that's what gets added when adding it from the website.

imZuxi:
> According to your website, shouldn't this be https://avtr.zuxi.dev/search? At least that's what gets added when adding it from the website.

either url works however avtr.icu goes over cloudflare with less restrictions as avtr.zuxi.dev

imZuxi:
> Unable to connect to `https://avtr.icu/search` due to anti-bot. By the way, do you the owner of the world or search provider?

i am the owner and thats super odd i dont have anti bot on err any type of blocking other than anti scrape

Misaka-L:
> > Unable to connect to `https://avtr.icu/search` due to anti-bot. By the way, do you the owner of the world or search provider?
> 
> i am the owner and thats super odd i dont have anti bot on err any type of blocking other than anti scrape

maybe because i'm using VPN and have IP with a low trust score... nerver mind

osiris-plus:
Could open a window with webview to have user pass cloudflare challenge. Then store the cookies for search requests.

imZuxi:
> Could open a window with webview to have user pass cloudflare challenge. Then store the cookies for search requests.

its very possible however i dont know if that would be allowed nor would i thunk it be required as i do have some whitelist rules directly for VRCX

osiris-plus:
wdym won't be allowed? You are a real user passing a security challenge to use the service. I don't see anything malicious here. This is how imageboard clients get around when user wants to post something.

imZuxi:
> wdym won't be allowed? You are a real user passing a security challenge to use the service. I don't see anything malicious here. This is how imageboard clients get around when user wants to post something.

ultimately it be up to @Natsumi-sama or other core maintainers to decide if its a feature that they would like to implement same with accepting this pull request

alexderpyyfox:
Is this as replacement or just as extra? Because ICU is quite bad and doesn't allow searches with only 3 letters ( as examples you can't type in bee or fox) it requires 4 at minimum overal it doesn't give good results

imZuxi:
> Is this as replacement or just as extra? Because ICU is quite bad and doesn't allow searches with only 3 letters ( as examples you can't type in bee or fox) it requires 4 at minimum overal it doesn't give good results 

this is a extra search it uses a completly different algorithm to be as fast as possible and is still being tweaked. as far as the minimum thats a weird bug as the limit has been reset to 2 letters a bit ago

PR #1642, opened 2026-02-12
Correct the incorrect translation - Update ja.json

neco222:
Correcting Japanese mistakes
時→時間


neco222:
I revised the server status translation because it felt awkward.
It now displays in a way that is easier for Japanese people to understand accurately.

PR #1654, opened 2026-02-20
Fixed overflowing text in a cell drawing over the adjacent cell

osiris-plus:
Fixed overflowing text in a cell drawing over the adjacent cell.

Before:
image

After:
image


Map1en:
I doubt this case could be reproduced

osiris-plus:
> I doubt this case could be reproduced

In Feed tab change the size of User column to the minimum.

VRCX version 2026.02.11

PR #1663, opened 2026-02-28
Ensure log directory exists

zmo2s:
### Summary
This PR ensures that the log directory exists, refactor string literal to const variable, class to record.


Natsumi-sama:
I found no bug with logs folder not being created from my testing.

PR #1678, opened 2026-03-07
Adds a configurable webhook event export system for VRCX

Sakura1618:
## Summary

Adds a configurable webhook event export system for VRCX.

This PR introduces a centralized webhook event registry, event-level enable/disable controls, and UI support for managing exported webhook events. It also adds runtime suppression for a few noisy initialization-phase events to reduce webhook spam after login.

## Changes

- add centralized webhook event registry and normalization utilities
- add per-event webhook export filtering in runtime delivery flow
- persist webhook enabled-event map in settings store
- add dedicated webhook event filter dialog in Settings
- wire webhook event export into notifications, favorites, and user-related event producers
- suppress selected high-frequency init-phase events during initial load
- add/update tests for registry, settings, UI, and event helper behavior
- set all webhook events to disabled by default

## Included Event Areas

- app lifecycle
- VRChat lifecycle
- self/location events
- favorites
- friend state updates
- instance metrics
- notification events

## Spam Reduction

To reduce noisy startup bursts, this PR suppresses these events before initial friends/favorites loading completes:

- `favorite.avatar.added`
- `friend.time_together_updated`
- `friend.note_updated`
- `friend.avatar_changed`

## Default Behavior

All webhook events are now **disabled by default** and must be explicitly enabled by the user.

## UI

- keeps webhook settings in the existing Notifications settings panel
- moves event-level webhook filters into a dedicated dialog
- supports per-event toggle, bulk enable/disable, and reset to defaults

## Notes

- unknown or disabled webhook events are blocked at export time
- supports both static event keys and dynamic notification event rules
- removed the init gate test script from this branch as requested

## Testing

- added/updated tests covering:
  - webhook event registry
  - webhook export filtering
  - settings persistence
  - notification helper logic
  - settings UI / dialog behavior

## Files of Interest

- `src/shared/constants/webhookEvents.js`
- `src/service/webhookEvent.js`
- `src/service/webhookEventGate.js`
- `src/stores/settings/notifications.js`
- `src/stores/notification/index.js`
- `src/stores/notification/webhookEventHelpers.js`
- `src/stores/favorite.js`
- `src/stores/user.js`
- `src/views/Settings/dialogs/WebhookEventFiltersDialog.vue`


Map1en:
Thanks for the contribution and for the work that went into this PR. I appreciate the effort. However, after reviewing the changes and the overall design, I have several concerns.

First, this PR is very large (around 1200 lines) and introduces a significant new feature. For changes of this scope it would be better to open an issue first to discuss the design and scope before implementation. Large feature PRs without prior discussion are difficult to review and most projects typically cannot accept them directly.

Second, the current implementation inserts webhook emission calls into many specific methods across multiple stores (favorites, users, notifications, etc.). This is a highly invasive pattern that tightly couples webhook behavior with many parts of the codebase. Even though the delivery is wrapped in a helper, the emission points themselves are still scattered across business logic, which makes the integration fragile and harder to maintain. It also increases the chance that future changes will forget to emit events.

Related to this, while the PR introduces a webhook event registry, it does not really introduce a centralized internal event system. The registry mainly defines event names and filtering rules, but event emission is still handled directly inside domain stores. Ideally this kind of feature should be built around a more centralized mechanism (for example around Pinia actions/state subscriptions or a dedicated internal event layer) instead of modifying many individual methods across the project.

More importantly, webhook exports effectively become a public integration interface once users start relying on them. That means the event model needs careful design before implementation, including things like event naming conventions, event schemas, payload stability, versioning, and backward compatibility. At the moment these aspects are not clearly defined.

There are also other practical concerns that would need to be addressed for a webhook system, such as delivery semantics, retry behavior, rate limiting or throttling for high-frequency events, batching/queueing strategies, error handling, and configuration lifecycle. These design questions should ideally be discussed first before implementing a feature of this size.

Because of these concerns, I’m not able to merge this PR as it is. If you’re interested in pursuing this feature, please open an issue first so the design and scope can be discussed before any implementation work.

Thanks again for taking the time to contribute.

PR #1690, opened 2026-03-14
Bypass System Proxy checkbox

ACE-MiRacle:
## Description
This PR explicitly sets `UseProxy = false` and `Proxy = null` when no proxy is configured, preventing unintended use of system proxies.

## Changes
- Modified `Dotnet/WebApi.cs` to add else clause disabling proxy
- Modified `Dotnet/ImageCache.cs` to add else clause disabling proxy  
- Modified `Dotnet/Update.cs` to add else clause disabling proxy

## Why?
When `ProxySet` is false, the .NET HttpHandler may still use the system proxy by default. By explicitly setting `UseProxy = false`, we ensure that no proxy is used unless explicitly configured in VRCX.

## Testing
- Verified changes compile without errors
- Tested with proxy disabled - connections should bypass proxy
- Tested with proxy enabled - connections should use configured proxy

### DIFF
--- a/Dotnet/Update.cs
+++ b/Dotnet/Update.cs
@@ -26,15 +26,22 @@ namespace VRCX
         private static string AppImagePath = string.Empty;
         private static string AppImagePathOld = string.Empty;
 
-        static Update()
-        {
-            var httpClientHandler = new HttpClientHandler();
-            if (WebApi.ProxySet)
-                httpClientHandler.Proxy = WebApi.Proxy;
-
-            httpClient = new HttpClient(httpClientHandler);
-            httpClient.DefaultRequestHeaders.Add("User-Agent", Program.Version);
-        }
+static Update()
+{
+    var httpClientHandler = new HttpClientHandler();
+    
+    if (WebApi.ProxySet)
+    {
+        httpClientHandler.Proxy = WebApi.Proxy;
+    }
+    else
+    {
+        httpClientHandler.UseProxy = false;
+        httpClientHandler.Proxy = null;
+    }
+    httpClient = new HttpClient(httpClientHandler);
+    httpClient.DefaultRequestHeaders.Add("User-Agent", Program.Version);
+}
 
         public void Init(string appImagePath = "")
         {


Natsumi-sama:
Issue with this approach is it wont change the browser behavior, Cef (Windows), Electron (Linux).
Those will also need to have proxy disabled as well, and I'm sure some users will complain about their global proxies no longer working with VRCX so it might need to be an option.

Dotnet/Update.cs
```diff
@@ -26,15 +26,22 @@ public class Update
...
+        httpClientHandler.Proxy = WebApi.Proxy;
+    }
+    else
+    {
+        httpClientHandler.UseProxy = false;
+        httpClientHandler.Proxy = null;
+    }
+    httpClient = new HttpClient(httpClientHandler);
+    httpClient.DefaultRequestHeaders.Add("User-Agent", Program.Version);
+}
```
BenjaminZehowlt: Indentation is entirely misaligned.

Dotnet/Update.cs
```diff
@@ -268,4 +275,4 @@ public static async Task CancelUpdate()
             }
         }
     }
-}
\ No newline at end of file
+}
```
BenjaminZehowlt: Why was the trailing new line removed?

Dotnet/WebApi.cs
```diff
@@ -75,27 +75,31 @@ public void Init()
...
+    }
+    else
+    {
+        _httpHandler.UseProxy = false;
+        _httpHandler.Proxy = null;
+    }
 
+    _httpClient = new HttpClient(_httpHandler);
+    _httpClient.DefaultRequestHeaders.Add("User-Agent", Program.Version);
+}
```
BenjaminZehowlt: Indentation is entirely misaligned.

ACE-MiRacle:
> Side note, I didn't mention originally as I forgot to. We should still respect the system proxy settings. If this were to be implemented, we'd recommend adding it as a "Bypass System Proxy" checkbox.

So are you guys planning to implement the feature, Including CEF/Electron yourselves? Or are you guys expecting me to update my PR with those changes?

BenjaminZehowlt:
We have no obligation to implement this; it's an open-source project. We implement what's in our current scope. You have been acting entitled as if we're obligated to this.

PR #1697, opened 2026-03-18
installer maintenance mode, AppData install option, and CI improvements

affectioned:
### Maintenance mode (existing installs)

When VRCX is already installed, the installer now opens a **maintenance page** instead of proceeding blindly. From here you can update in-place, relocate the installation, or uninstall.

> Auto-updates triggered from within VRCX pass `/UPDATE` to skip this page entirely.

----------

### Location picker (fresh installs)

New installs now prompt for an install location upfront:

-   **AppData** — no UAC prompt required
-   **Program Files** — UAC is only invoked when actually needed

Relocating to the same location you're already on is blocked with an explanatory message.

----------

### Uninstaller: optional user data removal

The uninstaller now includes an optional step to delete `%APPDATA%\VRCX`. To prevent accidents, this requires a **two-step confirmation** before anything is removed.

----------

### UX fix: deferred process kill prompt

The prompt to kill a running VRCX process is now shown **after** the user has chosen an action, rather than immediately on launch.

----------

## CI Workflow Improvements

The monolithic build workflow has been split into **separate per-platform workflows** (Windows, Linux, macOS) for faster, isolated testing. x64 and arm64 `dotnet` builds now run in parallel via matrix jobs. All GitHub Actions have been updated to their latest versions.

Screenshot 2026-03-18 180313

**Before** — one large combined workflow:

**After** — split, parallelized workflows:


affectioned:
I have resolved conflicts, also updated installer.nsi quite a while ago to be less code and complicated and more about the AppData migration.

Natsumi-sama:
Unless a new maintain coming along, I wont be merging this, too much risk without thorough testing.

PR #1700, opened 2026-03-20
feat: check for whether mutuals are enabled

acmdf:
This PR uses a slight edge case in the VRC api where if the user has mutuals disabled, it will return 0 mutual groups in the mutual counts endpoint no matter what. We can then check whether the user actually has 0 mutual groups, and if this is untrue we know that they simply have mutuals disabled, and it sets a flag on their id, which is visible in the friend list or as a message in the mutual friends tab.

### DIFF
--- a/src/views/FriendList/columns.jsx
+++ b/src/views/FriendList/columns.jsx
@@ -384,7 +384,15 @@ export const createColumns = ({
                     label: () => t('table.friendList.mutualFriends')
                 }),
             size: 120,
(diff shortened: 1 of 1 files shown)


Map1en:
We think this feature has a high chance of false positives and doesn' t provide much value to users. Also, the current changes to the mutual friends database structure would affect existing history markers. So unfortunately, we' re not able to merge this PR.

PR #1734, opened 2026-04-11
Friends Locations, Friends Modal Refactor

shinyflvre:
This refactor improves how statuses are displayed.

1. The correct web/mobile status is now shown in the Active sidebar. Previously, no status was shown until opening a profile, which caused confusion.
2. The profile modal no longer shows a yellow dot that required hovering to reveal the actual status. It now uses the new status circles, which directly show Active, Join Me, Ask Me, or Busy, matching the VRChat website.

This is a follow up improvement to the previous PR and makes the UI clearer and easier to use for end users.

Please review `status-icon.css`. As far as I can tell, there was no proper CSS for these icons besides the glow version, so it may need to be rewritten. I am not fully sure at this point.

I did not delete `--status-active` because it is still referenced in the overlay. I did not want to touch the overlay in this PR, so I will probably open a follow up later.

I use `ctx.state` for the updates, same as in the previous PR #1732.

Examples:
image

image

you may verify before merge, as i have not overlooked the complete workflow. ctx seems fine here however.
In my next PR i will look over the overlay and the other places where status-active is still referenced.

### DIFF
--- a/src/components/dialogs/UserDialog/UserSummaryHeader.vue
+++ b/src/components/dialogs/UserDialog/UserSummaryHeader.vue
@@ -35,7 +35,7 @@
                         <template #content>
                             <span>{{ getUserStateText(userDialog.ref) }}</span>
                         </template>
-                        <i class="x-user-status" :class="userStatusClass(userDialog.ref)"></i>
+                        <i class="x-user-status" :class="dialogStatusClass"></i>
                     </TooltipWrapper>
                     <template v-if="userDialog.previousDisplayNames.length > 0">
                         <TooltipWrapper side="bottom">
@@ -263,7 +263,7 @@
 
 <script setup>
     import { Apple, ChevronDown, IdCard, Image, Monitor, Shield, Smartphone, UserPlus, Users } from 'lucide-vue-next';
-    import { ref, watch } from 'vue';
+    import { computed, ref, watch } from 'vue';
     import { storeToRefs } from 'pinia';
     import { useI18n } from 'vue-i18n';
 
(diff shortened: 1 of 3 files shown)


shinyflvre:
Uhh we love package locks out of sync. should i do something on my end? it shouldnt have any issue with main

Natsumi-sama:
You can safely ignore all that, it isn't fully setup yet.

shinyflvre:
Alright! thanks for the answer!

Natsumi-sama:
Thanks for the PR, it sent me down quite the rabbit hole but you gave me a good starting point, I found the root cause and fixed it with this https://github.com/Natsumi-sama/VRCX/commit/201843958adb58de575b729167e3f521cea39429

shinyflvre:
That works in the Friends sidebar, but not in the Active tab within the Friends location. Your reimplementation and fix introduced a new bug.

Friends are now only shown as “green,” while all other statuses are no longer displayed, such as Ask Me, Join Me, and Busy. Even after opening profile cards, there is still no update. That is why I used `ctx` directly on startup to initialize those states.

That is also why we had:

```js
const dialogStatusClass = computed(() => {
    const r = userDialog.value?.ref;
    if (!r) return {};
    if (r.state === 'active') {
        if (r.status === 'join me') return { 'active-joinme': true };
        if (r.status === 'ask me') return { 'active-askme': true };
        if (r.status === 'busy') return { 'active-busy': true };
        return { active: true };
    }
    return userStatusClass(r);
});
```

I was not fully sure whether my approach was the right one, since I only reviewed a few classes in the whole project.
You may need to go over your changes again, but I am sure you will be able to reimplement it better than I did. :)

I will have bit of better knwoledge as soo nas i understand the whole core of VRCX give me a week ^^

PR #1753, opened 2026-05-06
fix(localization): update Japanese localization keys and values

puk06:
I checked all the keys based on en.json and implemented everything except the “WhatsNew” section.

I also corrected any parts of the existing translations that felt a bit off.

~~Since I felt that the “WhatsNew” section contained a mix of new and old items, I plan to address the “WhatsNew” section in a separate pull request. (The items are completely different from en.json, and it looks like ja.json is the correct one!)~~

Since this part won’t be used in the software update, I’ve decided not to address it.

### DIFF
--- a/src/localization/ja.json
+++ b/src/localization/ja.json
@@ -17,9 +17,16 @@
             "reset": "リセット",
             "view_details": "詳細を表示",
             "configure": "設定",
-            "refresh": "更新"
+            "refresh": "更新",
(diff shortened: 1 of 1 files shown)


puk06:
I forgot to include my name in the translator section. Is it okay if I submit a pull request just to add it?

Natsumi-sama:
Go for it.

PR #1767, opened 2026-05-14
Add missing migration

JantsoP:
When updating to 2026.05.03 build, favorites gets moved to Tools->Auto Status & Invites. However, there was never a migration to move the array from "Only Accept From" field to the new format which caused the array to become null. This caused all requests from favorites silently to be not accepted as there is no favorite groups to accept.


Natsumi-sama:
It's been long enough now that I don't think this is necessary, if anything users might complain when the feature suddenly starts working again.

PR #1771, opened 2026-05-21
update packages to fix some audit warnings, update solution to slnx, add vue-tsc so type validation works

redwyre:
* Add launch.json and tasks.json to run in VSCode
* Migrate `sln` to `slnx` - opening in vs2026 updates the `sln` to a newer version anyway, so might as well migrate to the new `slnx` standard which is a lot more concise. Rider, VSCode, VS2022, and VS2026 all can understand the format as of around early 2025 so anything reasonably up to date will be fine. The `sln` file isn't used in building so this doesn't affect anything other than editing.
* Change `VRCX-Cef.csproj` debugging type to portable so it can work with VSCode, works with everything.
* Added `typescript` and `tsc-vue` for type checking. The setup was there it just didn't work, so this makes it work for `.js` and adds support for `.vue` files. I started fixing some warnings but moved that to a separate PR.
* Set `.npmrc` to only allow packages older than 7 days (which is main window of supply chain attacks)
* Updated packages due to vulnerabilities. Very selective update, nothing newer than a week old except where indicated.
  * `showdown` (markdown renderer) has a moderate vulnerability but since the input is controlled it should be fine, but it's long since abandoned, used by `vue-showdown`
  * updated `tmp` and added override so it can't be added as an older version
  * `js-cookie` was updated in another commit, but I've added an override so it can't be added as an older version
* Updated `build-all.ps1`‎ to use absolute path when creating symbolic link because it gave me an error
* Fixed deprecated file warnings in `generate-third-party-licenses.js‎`
* Updated electron dotnet from 9 to 10, to consolidate dotnet versions on 10.0.8
* Added `install-linux-deps.sh` to install all linux dependencies (.deb, Ubuntu). Used to set up testing environment for now but can be expanded later.


Natsumi-sama:
- Leave out npm updates as it's easy to add malicious packages to `package-lock.json`, also we don't want to update Electron due to a Linux bug.
- Why remove? https://github.com/vrcx-team/VRCX/pull/1771/changes#diff-a5de3e5871ffcc383a2294845bd3df25d3eeff6c29ad46e3a396577c413bf357L11
- Why change? https://github.com/vrcx-team/VRCX/pull/1771/changes#diff-ed8b0302426ae825aac304e345368a78d33f1e9022bce605fb79d375cf8e40c0R39

redwyre:
> * Leave out npm updates as it's easy to add malicious packages to `package-lock.json`, also we don't want to update Electron due to a Linux bug.

Covered in updated description

> * Why remove? https://github.com/vrcx-team/VRCX/pull/1771/changes#diff-a5de3e5871ffcc383a2294845bd3df25d3eeff6c29ad46e3a396577c413bf357L11

It is duplicated two lines below

> * Why change? https://github.com/vrcx-team/VRCX/pull/1771/changes#diff-ed8b0302426ae825aac304e345368a78d33f1e9022bce605fb79d375cf8e40c0R39

Doesn't work in VSCode, works the same as full

PR #1804, opened 2026-06-27
Fixed database migration race condition

Nymmh:
Fixes a database race condition that could occur on first launch (source: it happened to me). Migrations could run before the tables were initialized, causing it to fail.

Migration process will now initializes the tables before applying migrations. All the migrations have also been moved to their own file to make the code look a bit better.


src/stores/vrcx.js
```diff
@@ -233,16 +234,7 @@ export const useVrcxStore = defineStore('Vrcx', () => {
...
-                await database.cleanLegendFromFriendLog(); // fix friendLog spammed with crap
-                await database.fixGameLogTraveling(); // fix bug with gameLog location being set as traveling
-                await database.fixNegativeGPS(); // fix GPS being a negative value due to VRCX bug with traveling
-                await database.fixBrokenLeaveEntries(); // fix user instance timer being higher than current user location timer
-                await database.fixBrokenGroupInvites(); // fix notification v2 in wrong table
-                await database.fixBrokenNotifications(); // fix notifications being null
-                await database.fixBrokenGroupChange(); // fix spam group left & name change
-                await database.fixCancelFriendRequestTypo(); // fix CancelFriendRequst typo
-                await database.fixBrokenGameLogDisplayNames(); // fix gameLog display names "DisplayName (userId)"
-                await database.upgradeDatabaseVersion(); // update database version
```
redwyre, line 245: I think it might be cleaner to just use `await database.upgradeDatabaseVersion()` as the entrypoint here and keep everything inside `database.js` so you don't need to jump back and forth between two files

Natsumi-sama: Might need to rearrange more so we don't have to run `initTables` twice.

Nymmh: Moved migration logic into `database.upgradeDatabaseVersion()`.
Removed duplicate index's.
App.vue will now wait for database init to prevent access during an incomplete migration.

redwyre (approved):

Issue #1541, opened 2025-12-20
[Feature Request] Add some EXIF data to saved stickers


M1XZG:
**Explain in detail what your suggested feature would be used for.**

For moderation (and perhaps generally) it would awesome if some EXIF data could be saved in the stickers that get saved when someone drops it in in a world.

**Describe how it would look if it requires a UI.**

Ideally it would be similar to the EXTRA data stored in photos that we take but perhaps not quite as extensive. A list of all users wouldn't be required

| Category | Field           |
|----------|-----------------|
| App      | Application     |
| App      | Version         |
| Author   | ID              |
| Author   | Display Name    |
| World    | Name            |
| World    | World ID        |
| World    | Instance ID     |
| World    | Instance Number |
| World    | Privacy         |
| World    | Owner (User ID) |
| World    | Region          |
| World    | Nonce           |

**Explain why people would want to use it.**

This would be especially useful I feel for moderators of instances where we may need to report a sticker placed by someone after the fact. This has come up a few times in a couple communities. The sticker wasn't able to get reported in-game before the person was kicked for example. So going back to ID the person that placed it would be extremely useful. Having the extra info would also help VRC in the reports.

PR #1808, opened 2026-06-28
feat: embed VRCX metadata in saved stickers (#1541)

VVarren:
Hi, 

This is my first time contributing to open-source. Please let me know if I should adjust anything in the PR.

PR based off #1541 

Note: 

- Instance ID had repetitive data so it wasn't added
- nonce was excluded because anyone who extracts the metadata from the png will have access to join that private instance

result


Natsumi-sama:
For this to be merged it needs to correctly follow existing VRCX metadata schema, no extra fields, instanceId as location. Emojis should also have the same metadata set as stickers.

redwyre:
I'm not sure this makes sense since a sticker/emoji isn't related to a single place, it will only be the first place you see it. Eg. what happens the second time you see the sticker in a different world, do you want to save it again with new metadata?

PR #1829, opened 2026-07-24
Fix: show Add  Add Favorite button not shown for non-friend usersFavorite button for non-friend users

SpacyRainbow:
The favorite star button in the user profile action dropdown
(UserActionDropdown.vue) only rendered when the target was already a
friend or already a local favorite:

    v-if="(currentUser.id !== userDialog.ref.id && userDialog.isFriend) || userDialog.isFavorite"

This meant there was no way to add a non-friend to a local favorite
group from the UI, even though the underlying favorite dialog and the
local-favorites store never required friendship in the first place —
local favorites are meant to work for non-friends (e.g. tracking
join/leave notifications for someone you're not friends with).

Fix: the button now only excludes the current user themself.

    v-if="currentUser.id !== userDialog.ref.id"

Added two tests confirming the button renders for a non-friend/
non-favorite user and that clicking it still sends the Add Favorite
command correctly.

### DIFF
--- a/src/components/dialogs/UserDialog/UserActionDropdown.vue
+++ b/src/components/dialogs/UserDialog/UserActionDropdown.vue
@@ -1,6 +1,6 @@
 <template>
     <div style="flex: none" class="flex items-center">
-        <template v-if="(currentUser.id !== userDialog.ref.id && userDialog.isFriend) || userDialog.isFavorite">
+        <template v-if="currentUser.id !== userDialog.ref.id">
             <TooltipWrapper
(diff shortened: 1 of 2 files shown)


Natsumi-sama:
This may not "just work" like you'd expect, but there little harm in adding it.

SpacyRainbow:
I thought this was unintended behavior. I understand now that this can be used maliciously

Natsumi-sama:
PR doesn't work, please test changes before creating PRs.

PR #1837, opened 2026-08-01
Enable "Open in Game" on Linux

different-name:
Currently on Linux the "Open in Game" functionality is disabled. This PR steps into the running game's namespaces (nsenter) and hands the deeplink straight to it with `wine launch.exe ...&attach=1`, making the open in game button it work the way it does on Windows

If proton isn't being used, no nsenter, or the game isn't running, it quietly falls back to the self-invite like before. The "Self Invite Action" setting can also be used to opt-out like on Windows

### DIFF
--- a/Dotnet/AppApi/Common/AppApiCommon.cs
+++ b/Dotnet/AppApi/Common/AppApiCommon.cs
@@ -169,7 +169,7 @@ namespace VRCX
             return null;
         }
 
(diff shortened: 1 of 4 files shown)


konsti219 (approved):
Tested on my system and it works

Natsumi-sama:
This looks awful, thanks, ship.exe

PR #1848, opened 2026-08-10
fix: expand active friends group by default

AariyJP:


AariyJP:
Every other friend group defaults to expanded — is collapsing only ACTIVE intentional?

Natsumi-sama:
This line is only for users first installing VRCX.

It was originally to help with performance, but I think friends are dynamically unloaded now.

AariyJP:
Since VRChat players relatively have high-spec PCs, I think setting it to true is fine.
Also, suppose you wanted to improve performance, don't you think it's unnatural for only ACTIVE to be false?

Natsumi-sama:
PC specs isn't the issue, the browser can easily lag and crash from having too much data loaded.
The active list isn't very useful most of the time and users can always enable it when needed.

Natsumi-sama:
Looking at this part of the code again, it has changed a lot since I last touched it. Sidebar is a lot more performant now and it wont suffer from these issues anymore. Offline friends is already defaulted to enabled, active should be enabled as well.

PR #1860, opened 2026-08-16
fix: show queue size indicator

AariyJP:
`hasInstanceMetadata` reads `props.instance.value?.queueSize`, but `props.instance` is a plain props object rather than a ref, so `.value` is always `undefined`.
The queue term never opens the gate, and instances whose only metadata is a queue render no indicator at all.


Natsumi-sama:
Where does this bug materialize? in sidebar group instances?

AariyJP:
The queue indicator goes missing in
* User dialog — the instance chip under a user's current location
* Group dialog → Info tab instance rows
* World dialog → Instances tab instance rows

PR #1864, opened 2026-08-18
Copy external links with Ctrl+C

### LINKED ISSUE #1859: [Feature Request] ctrl-c copy "Open external link" dialog to copy
**Explain in detail what your suggested feature would be used for.**
It would be handy to be able to hit ctrl-c when the "Open External Link" modal is open to have it auto close and copy to clipboard (same as the cancel/copy action).

**Describe how it would look if it requires a UI.**
It does not require a visual UI change. Ideally doing so does the same thing as the "Copy" action as-is - close the modal, copy to clipboard, show the "Link copied to clipboard!" toast.

**Explain why people would want to use it.**
Because I run vrchat over parsec and don't log into websites on Windows (even vrchat's own site), being able to more quickly copy links to stuff when the prompt opens would be a nice workflow tidbit.

markoapple:
Fixes #1859.

The Open External Link dialog currently requires clicking Copy when the user wants the link in the clipboard, which makes keyboard-driven and remote workflows slower.

Move the external-link confirmation into its own dialog. Ctrl/Cmd+C uses the same copy action as the Copy button, while built-in VRCX links continue to use direct-access routing without showing the dialog.

Tests:

* `npm test -- src/shared/utils/__tests__/appActions.test.js src/stores/__tests__/modal.test.js src/components/__tests__/OpenExternalLinkDialog.test.js` (40 passed)
* `npx oxlint` and `npx eslint` on the touched files
* `npx oxfmt --check` on the new files
* `git diff --check`
* `npx vite build src`


Natsumi-sama:
I would much rather the copy alert be its own modal, rather than adding copy related code to the general alert modal.

markoapple:
Moved this to a dedicated modal and removed the generic alert changes. Thanks.

PR #1883, opened 2026-09-16
Add Linux desktop and AppStream metadata for Linux package maintainers

nexryai:
## Summary

Add Linux desktop integration metadata to the repository:

* Add a `.desktop` file for application launchers and desktop environments
* Add AppStream metainfo for software centers and package managers

These files can be reused by Linux distribution packages instead of maintaining application metadata separately in downstream packaging repositories.

## Motivation

VRCX already includes some Linux desktop metadata in the Electron Builder configuration, but does not currently provide standalone desktop or AppStream metadata files.

Providing these files upstream makes the application's Linux integration metadata available to downstream packagers and avoids duplicating it in packaging repositories.

This also addresses feedback from the Flathub submission review requesting that the desktop and metainfo files be maintained upstream.

https://github.com/flathub/flathub/pull/10231

https://github.com/flathub/flathub/pull/10231#discussion_r4017079121

https://github.com/flathub/flathub/pull/10231#discussion_r4017109591


redwyre:
I was planning on adding AppStream data for the AppImage eventually so I'm fine with that, though I'm not so hot on duplicating the desktop file. As a side note you are able to build flatpak directly from electron-builder: https://www.electron.build/docs/flatpak

There are a number of concerns here:
* currently the app uses `app.vrcx`, but I have been looking at updating that to `app.vrcx.vrcx`, so for official appstream  we would use that
* duplicating `desktop` file, it's saved out as part of the appimage building with electron-build, you could use the flatpak building of electron-build, or extract it from the js file (should be exported at something like `linux.desktop.entry`)
* I have concerns of using "official" id in unofficial packages as people may think they are official and supported, but I don't know a much about this area

PR #1898, opened 2026-09-20
Go back one crumb instead of closing dialog, typo fix

GapVR:
When clicking outside a dialog with crumb trail, go back one crumb instead of closing the dialog box immediately.

Also fixed a typo in english localization.


Natsumi-sama:
This dialog behavior is well known and intended, I think changing it isn't a good idea.

DorCoMaNdO:
> This dialog behavior is well known and intended, I think changing it isn't a good idea.

Back when it was introduced one of my suggestions was to add a setting to close (current behavior)/go back (better matching the previous behavior with layered dialogs)/do nothing when clicking the background, maybe repurpose this PR for that?

PR #1901, opened 2026-09-21
fix(api): throttle calendar requests to respect VRChat API rate limit

### LINKED ISSUE #1900: Calendar frequently fails with 500/429 — VRCX doesn't throttle VRChat API requests
# Calendar frequently fails with 500/429 — VRCX doesn't throttle VRChat API requests

## Summary

The Group Calendar (and other calendar features) frequently shows a red error toast like the following when navigating between months or rapidly opening event details:

```
500 Internal Server Error
Path: /api/1/calendar
Error Message: Application error
Endpoint: "calendar"
Status Code: 500
```

The calendar UI still renders — day cells and the last-known event badges stick around from a previous successful load — but the events for the visible month may be missing or stale. This has been the case for months across many users.

## Repro

1. Open VRCX and log in.
2. Open a group that has an active calendar (or the global Calendar view).
3. Click the **Next Month** / **Previous Month** buttons 3–5 times in a few seconds, or rapidly open event detail dialogs.
4. Observe a `500 Internal Server Error / Endpoint: "calendar"` toast.
5. Reload the calendar: events may be missing or stale.

It also reproduces more reliably if you have many friends + active group instances + the calendar open — anything that already pushes VRCX close to the VRChat rate limit.

## Expected

- Calendar events load reliably across month navigation.
- Either VRCX respects VRChat's rate limit, or a 429/500 response surfaces with a clear "rate limited, please slow down" message rather than a generic "Application error".

## Actual

- Some or all calendar events fail to load after rapid navigation.
- A generic 500 toast is shown with no actionable information.
- The calendar UI keeps the last-known event map until a fresh fetch replaces it — partial data sticks around even when the new request 500s, which makes the failure look intermittent.

## Root cause

VRChat's REST API has a combined rate limit of roughly **60 requests per minute across all endpoints** (not per-endpoint). When you navigate the calendar quickly, VRCX fires multiple calendar requests in rapid succession:

- `GET calendar/{groupId}` — group calendar (once per refresh / group switch)
- `GET calendar/{groupId}/{eventId}` — per event detail (each dialog open)
- `GET calendar` — current user's calendar list
- `GET calendar/following` — calendars of groups the user follows
- `GET calendar/featured` — featured calendars

Combined with the rest of VRCX (friend status polling, instance refreshes, group updates, friend requests) a single fast user can blow past 60 reqs/min within seconds. VRChat responds with **429** — or, when the load is high, **500** from the edge — and VRCX surfaces it as a generic error toast.

## What I checked in the code

Walked the call graph with CodeGraph end-to-end:

```
getGroupCalendar(groupId)                                  src/api/group.js
  └─ request('calendar/{groupId}')                         src/services/request.js:87
      └─ webApiService.execute({ url: '…/calendar/{groupId}', method: 'GET' })
                                                            src/services/webapi.js:20
          └─ WebApi.Execute(options)                        Dotnet/WebApi.cs:384
              └─ _httpClient.SendAsync(request)             (no rate-limit wrapper)
```

Findings:

1. **Vue side (`src/services/request.js`)** deduplicates identical in-flight GETs for 10 seconds, but does **not** throttle distinct requests. There is no rate limiter on the calendar path.
2. **C# side (`Dotnet/WebApi.cs`)** uses `HttpClient` with `MaxConnectionsPerServer = 10` — that's connection-pool sizing, not request throttling. No rate-limit logic exists in the .NET project (verified by `codegraph query rateLimit|throttle|debounce -p Dotnet` → 0 results).
3. **`createRateLimiter`** already exists at `src/shared/utils/throttle.js` and is used by `fetchMutualGraph` (charts) and `bulkRefreshFriends` (friends) — but **not** by anything in the calendar path. Five total callers; calendar is not one of them.
4. The 429 handler in `request.js` only fires for `…/instances/groups` (sets a 120s backoff on `setNextGroupInstanceRefresh`). 429s from any other endpoint, including calendar, fall through with no specific handling.

## Suggested fix

1. Wrap calendar requests with the existing `createRateLimiter` (e.g. **12 requests / 60 seconds** — well under the combined 60 reqs/min VRChat limit, leaving headroom for friends / instances / group updates). Same pattern as `fetchMutualGraph` and `bulkRefreshFriends`.
2. Surface 429s from any endpoint as proper status-coded errors instead of letting them fall through to the generic 500 path.
3. (Optional) On the C# side, surface the `Retry-After` header from VRChat's 429 response so the frontend can compute a precise backoff.

## Environment

- VRCX 2025.12.06 (and recent builds — this has been the case for months)
- VRChat API rate limit: ~60 reqs/min combined
- Reproduced on Windows; likely affects Linux Electron build as well

## Severity

Medium — the feature is degraded but not unusable (stale data is still shown), and most casual users won't hit it. Power users who navigate the calendar fast or have many friends are affected regularly.

ComfyChloe:
# fix(api): throttle calendar requests to respect VRChat API rate limit

## Summary

Adds a sliding-window rate limiter to the `calendar/*` endpoint family in `src/services/request.js` so VRCX stops blowing past VRChat's combined ~60 reqs/min API limit when users navigate the calendar quickly. Also surfaces 429 responses from any endpoint as proper status-coded errors instead of letting them fall through to the generic "Application error" toast.

Closes #1900

### `src/services/request.js`

1. **Imported** the existing `createRateLimiter` from `src/shared/utils/throttle.js`.
2. **Added** a `calendarRateLimiter` instance at **12 reqs / 60 seconds** — well under VRChat's combined 60 reqs/min limit, leaving headroom for friends / instances / group updates so we don't just shift the problem.
3. **Applied** the limiter to all `GET` requests whose endpoint starts with `calendar`, *before* dispatching to `webApiService.execute`. The limiter is awaited via `.then()` so the function still returns the same Promise shape — no async/await change needed and no breakage for existing callers.
4. **Removed** the `init.url.endsWith('/instances/groups')` gate on the existing 429 handler, so 429s from any endpoint surface as proper status-coded errors (the calendar case in particular was previously falling through silently after the `parseError` branch).

```js
// new
const calendarRateLimiter = createRateLimiter({
    limitPerInterval: 12,
    intervalMs: 60_000
});

// wrapped dispatch
req = (init.method === 'GET' && endpoint.startsWith('calendar')
    ? calendarRateLimiter.wait().then(() => webApiService.execute(init))
    : webApiService.execute(init))
    .catch((err) => { $throw(0, err, endpoint); })
    // …rest unchanged…
```

```js
// 429 handling — was scoped to /instances/groups, now applies to all endpoints
if (parsed.status === 429) {
    if (init.url.endsWith('/instances/groups')) {
        updateLoopStore.setNextGroupInstanceRefresh(120); // 1min
    }
    $throw(429, t('api.status_code.429'), endpoint);
}
```

## Why this works

- **`createRateLimiter`** is the same utility already used by `fetchMutualGraph` (`src/stores/charts.js:235`) and `bulkRefreshFriends` (`src/stores/friend.js:658`) — proven pattern in the codebase. No new abstractions.
- **12 reqs/min is conservative.** It leaves ~48 reqs/min for the rest of VRCX (friends, instances, group updates), so this fix doesn't just push the limit-hit to another subsystem.
- **Per-process limiter** matches how the existing rate limiters in the codebase work — there's no cross-session coordination needed because each VRCX process talks to VRChat with its own cookies.
- **`.then()` wrapping** keeps `request()` synchronously returning a Promise — no async/await refactor, no breakage for any of the ~20 callers of `request()`.

## Why not change `Dotnet/WebApi.cs`?

The C# `WebApi.Execute` already passes the upstream HTTP status code through correctly — it returns a `Tuple<int, string>` from the `HttpClient` round-trip, and `WebApi.ExecuteJson` serialises it as `{ status, message }`. The 500 the user sees in the toast is what VRChat itself returned to VRCX (rate-limit edges sometimes return 500 instead of 429); it isn't VRCX masking a 429 as a 500.

A C#-side rate limiter would also work, but the Vue-side fix is smaller, scoped to just the calendar path, and matches the codebase pattern. Adding it in two places would just be more code to maintain.

## Risk / backwards compatibility

- **Behavioural change:** users who previously hit 500 toasts after rapid calendar navigation will now see the requests pace out cleanly. No more 500s on calendar navigation. If the upstream is *still* rate-limited (e.g. heavy parallel load from many VRCX subsystems), 429s will surface as proper 429 toasts — clearer than the previous generic 500.
- **No API change.** The shape of `request()` is unchanged. All ~20 callers keep working.
- **No new dependencies.** `createRateLimiter` already lived in `src/shared/utils/throttle.js`.

## Testing

- [ ] `pnpm lint` — no new warnings.
- [ ] `pnpm test` — existing tests still pass (no new tests added; the change is small and existing `throttle.test.js` covers `createRateLimiter` itself).
- [ ] **Manual repro of the original bug:** open a group calendar, click Next/Previous month 5+ times rapidly → no 500 toast, all events load.
- [ ] **Manual sanity:** friends list, instance refresh, group dialogs all still work normally (limiter is scoped to `calendar/*` only).

## Out of scope (filed as follow-ups)

- Surfacing `Retry-After` from VRChat's 429 response on the C# side (`WebApi.cs`).
- A global / process-wide request limiter that throttles all endpoints combined (would address the same root cause more robustly, but is a bigger change).

## Checklist

- [x] Code follows the existing `createRateLimiter` pattern.
- [x] Commit message explains the why.
- [x] Drafts of issue + PR live in `docs/drafts/` for review before posting.


ComfyChloe:
Don't let wild sheep into your fields, They might start growing mushrooms that make you see things..

PR #1902, opened 2026-09-21
Show last known location

MLFK-MLFK:
Following up on the feedback in the issue — I've simplified quite a lot and
removed a fair amount of dead code. Everything has been verified locally.
 I'd still appreciate another look at the code before agreeing to merge the
PR.

----------------------------------

## What this does
﻿
When a friend sets their status to Ask Me, Do Not Disturb, or otherwise goes
private, their instance is no longer visible. They often haven't actually left
the instance they were just in, so this adds a small section to the "Current 
instance" block in the user dialog showing the last instance they were recorded
in, with the usual join / invite actions.
﻿
It's shown only while their own location is hidden, and is hidden again once the
recorded instance is older than a configurable max age (default 1 hour),

----------------------------------

##UI modification reminder

Added inside the existing "Current instance" card, separated by a dashed rule so
it reads as additional information rather than a replacement:
﻿
- instance name (click opens the world dialog)
- access type and instance id, via the existing `LocationWorld` component
- join / invite buttons, via the existing `InstanceActionBar` component
- how long ago that instance was current
﻿
The max age is configurable under Settings → Interface → User dialog, where the
feature can also be turned off entirely.


OctoMiku01:
Don't think this is a good idea. Someone switching their status to orange/red is doing so for a reason. They do not want people to come in unannounced without asking or don't want anyone of their friends at all joining. This would just make people learn to switch their status before they are joining new instances or unfriend people that suddenly join via tools like this.

MLFK-MLFK:
> Don't think this is a good idea. Someone switching their status to orange/red is doing so for a reason. They do not want people to come in unannounced without asking or don't want anyone of their friends at all joining. This would just make people learn to switch their status before they are joining new instances or unfriend people that suddenly join via tools like this.

Indeed, you are right, this is indeed a controversial topic. But what I think is that even without this feature, the recorded historical data (GPS) alone can still replicate this behavior. Or perhaps the function of constantly recording historical data itself may already be somewhat controversial. This feature was originally not intended to be used in this way. Instead, it is used for situations such as invitation records, or when the player disconnects for some reason during gameplay, or temporarily goes to another place and then returns, or when the invitation expires. This feature allows for quickly finding the last existing room. I personally think that its harm may not be as high as imagined. However, I still strongly agree with your viewpoint. If possible, I can close the pull request (PR), but I think it's a bit harsh, because it's just a simplified version of a function that can already be achieved in the original version. People with ulterior motives will still not be deterred from illegally entering. On the contrary, the emergence of this feature will remind them. I feel that people who need to hide themselves should have the "right to know" about this feature

MLFK-MLFK:
> Don't think this is a good idea. Someone switching their status to orange/red is doing so for a reason. They do not want people to come in unannounced without asking or don't want anyone of their friends at all joining. This would just make people learn to switch their status before they are joining new instances or unfriend people that suddenly join via tools like this.

In my personal opinion, rather than concealing it, it might be a better choice to make it public
open to your opinion on whether it's appropriate

